PR #44, opened 2025-09-10
fix: resolve async streaming error in RFE agent analysis

jeremyeder:
## Summary
- Fix async streaming error that was preventing RFE agent analysis from completing
- Remove incorrect `await` from `astream_structured_predict` calls in streaming functions
- Add comprehensive error handling and fallback responses

## Problem
The RFE builder was failing with error: `'async for' requires an object with __aiter__ method, got coroutine`

## Root Cause
The streaming functions were incorrectly using `await` on `Settings.llm.astream_structured_predict()` which returns an async generator, not a coroutine.

## Solution
- Removed incorrect `await` from streaming helper functions
- Added proper error handling to gracefully handle streaming failures
- Added fallback responses when streaming doesn't work properly
- Formatted code with black

## Test Plan
- [x] Verified RFE builder no longer shows async error
- [x] Confirmed agent analysis completes successfully
- [x] Code passes black formatting

### DIFF
--- a/demos/rfe-builder/src/agents.py
+++ b/demos/rfe-builder/src/agents.py
@@ -19,29 +19,34 @@ async def stream_structured_predict(
     output_cls, prompt_template, persona: str, **prompt_args
 ):
     """Simple streaming with UI events every 50 chars"""
-    stream_generator = await Settings.llm.astream_structured_predict(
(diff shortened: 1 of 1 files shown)


jeremyeder:
tested locally and works

PR #68, opened 2025-09-11
migrate https://github.com/Gkrumbach07/claude-runner to vTeam demos

Gkrumbach07:
Moving the claude runner codebase into vTeam as a demo, should be intrgated in alter step as the core app


Gkrumbach07:
aprpving this, just merging efforts into the vteam

/approve
/lgtm

PR #81, opened 2025-09-16
Update README with comprehensive private registry authentication guide

natifridman:
- Add comprehensive private registry authentication guide
- Include detailed ImagePullBackOff troubleshooting for container registries
- Fix sed command for macOS compatibility (use sed -i '' instead of sed -i)
- Add service account patching instructions for image pull secrets


cooktheryan:
we shouldn't need to auth for private images. If we have any images that need auth then we should move those public

natifridman:
I agree. This PR was initially created to solve some other things in the README that are already part of #79 
Closing this PR

PR #86, opened 2025-09-17
adds git config and spec-kit to claude-code-runner

sallyom:
Adds spec-kit and git configuration to claude-code-runner


Gkrumbach07:
this looks good, will follow up with fixes in following PRs

PR #87, opened 2025-09-18
rename of file and add deploy action

cooktheryan:


cooktheryan:
Self merging due to it being a github action

PR #101, opened 2025-09-22
copy secrets from parent to child project

sallyom:
AFAICT we need to copy the runner-secret from parent to child 
This PR
1) adds secret copy from parent to child project in operator
2) adds ability to set namespace in kustomization.yaml that will be honored in deploy.sh
3) fixes indentation in deployment yaml
4) removes some unnecessary files 

THIS IS NOT THE WAY it will be - we'll add reconcilers in components/operator - I added to main.go to see it working quickly - not gonna merge like this


sallyom:
closing for now, will pick back up later - focusing on a few things b4 this!

PR #106, opened 2025-09-23
jira integration updates

sallyom:
AFAICT with this PR:
* Can upload and resync to Jira from `specify`, `plan`, and `tasks` but don't upload jira from `ideate` phase
* Can now delete RFE-workload from UI
* `ideate` phase optional - if no `rfe.md` then specify will use the RFE description instead.

### DIFF
--- a/components/backend/handlers.go
+++ b/components/backend/handlers.go
@@ -754,7 +754,7 @@ func provisionRunnerTokenForSession(c *gin.Context, reqK8s *kubernetes.Clientset
 			{
 				APIGroups: []string{"vteam.ambient-code"},
 				Resources: []string{"agenticsessions"},
-				Verbs:     []string{"get", "list", "watch"},
+				Verbs:     []string{"get", "list", "watch", "update", "patch"},
 			},
 			{
 				APIGroups: []string{"vteam.ambient-code"},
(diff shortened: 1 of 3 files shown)


components/backend/handlers.go
```diff
@@ -3430,13 +3440,18 @@ func publishWorkflowFileToJira(c *gin.Context) {
 	if existingKey == "" {
 		// Create
 		jiraEndpoint := fmt.Sprintf("%s/rest/api/2/issue", jiraBase)
+		// Determine issue type based on file type
+		issueType := "Feature"
+		if strings.Contains(req.Path, "plan.md") {
+			issueType = "Feature"  // plan.md creates Features for now (was Epic)
+		}
+
```
sallyom, line 3448: will change the plan.md to use the issueType "Epic" in future

components/backend/handlers.go
```diff
@@ -3446,15 +3461,15 @@ func publishWorkflowFileToJira(c *gin.Context) {
...
 				"description": content,
 			},
 		}
 		payload, _ := json.Marshal(reqBody)
 		httpReq, _ = http.NewRequest("PUT", jiraEndpoint, bytes.NewReader(payload))
 	}
 	httpReq.Header.Set("Content-Type", "application/json")
-	httpReq.Header.Set("Authorization", "Basic "+base64.StdEncoding.EncodeToString([]byte(jiraEmail+":"+jiraToken)))
+	httpReq.Header.Set("Authorization", "Bearer "+jiraToken)
 	httpClient := &http.Client{Timeout: 30 * time.Second}
```
sallyom, line 3473: we'll need both of these probably? cloud jira uses basic I think?  but for now , this works with our test Jira.

PR #145, opened 2025-10-06
git auth add ability to authenticate without GH application at project-level

sallyom:
* Added to the runner secret Git Auth - no SSH key, only API-TOKEN
* Runner will use either GH Application auth and/or GH_TOKEN in secret to authenticate for git operations (push or when working with private repos)
* Updated the runner secret to add a default ANTHROPIC_API_KEY field as well as the Git fields


sallyom:
@Gkrumbach07 ok to merge this? 
@bobbravo2 this should allow for git auth with local CRC setup & runner secret

bobbravo2:
>  @bobbravo2 this should allow for git auth with local CRC setup & runner secret

Thank you 🙏

PR #156, opened 2025-10-09
Fix token and other fixes

sallyom:
@Gkrumbach07 I took your branch and fixed things on top of it! PTAL 
Can test with default image names and repo `quay.io/sallyom`


sallyom:
merged into https://github.com/Gkrumbach07/vTeam/pull/4 instead!

PR #159, opened 2025-10-09
fix deploy.sh, remove .env

sallyom:
make .env file optional


sallyom:
@bobbravo2 does the local setup use `.env` file to deploy with?

bobbravo2:
> @bobbravo2 does the local setup use `.env` file to deploy with?

Looks like we are here: https://github.com/ambient-code/vTeam/blob/main/components/scripts/local-dev/crc-start.sh#L38-L46

Although the script should also help set up a new one, and it's mainly for `CRC_*` things (memory, cpu, etc)

I'm not _**super**_ worried about a breaking change here, esp if we're seeing more data/engagement from the Open Shift hosted solutions.

PR #173, opened 2025-10-12
remove duplicate AutoPushOnComplete

sallyom:
@Gkrumbach07 with the rebase I had this defined in 2 places - it should _only_ be [here](https://github.com/ambient-code/vTeam/blob/main/components/backend/types/session.go#L74) - the duplicate was causing issues - I hope this fixes it, testing now.

### DIFF
--- a/components/backend/git/operations.go
+++ b/components/backend/git/operations.go
@@ -36,13 +36,10 @@ type ProjectSettings struct {
 	RunnerSecret string
 }
 
-// DiffSummary holds summary counts from git status --porcelain
+// DiffSummary holds summary counts from git diff --numstat
 type DiffSummary struct {
-	Added     int `json:"added"`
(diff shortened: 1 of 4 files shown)


sallyom:
@cooktheryan @Gkrumbach07 
I've tested this and the auto-push is fixed - however I needed to build the new operator image. 
Maybe `quay.io/ambient_code/vteam_operator:latest` isn't up-to-date with latest operator code? 
I have confirmed from testing that `quay.io/ambient_code/vteam_operator:latest` does not set the AUTO_PUSH_ON_COMPLETE env var in the agenticsession jobs. Only when I build the operator from latest code does that get set. However, I'm not changing any operator code here, so the ambient_code/vteam_operator:latest must not have the latest code.

sallyom:
merging this fix!

PR #178, opened 2025-10-13
fix the issue where sessions fail for a few seconds before succeediing

sallyom:
Remove the check for empty result_payload upon completion - empty result_payload isn't necessarily a failure. 
@Gkrumbach07 ptal!


Gkrumbach07:
looks good

PR #184, opened 2025-10-14
add agent  to ui

MichaelClifford:
Simple PR to make vTeam agents more prominent in the UI. Added a page that lists the available agents from the vTeam project, along with some simple instructions for how to use them directly during an interactive session.


MichaelClifford:
Closing in favor of #185

PR #193, opened 2025-10-17
Add logout option to user menu

bobbravo2:
## Summary
- Added logout menu item to user dropdown menu with LogOut icon
- Implements cookie clearing functionality to log out users
- Redirects to home page after logout

## Test plan
- [ ] Verify logout option appears in user dropdown menu
- [ ] Click logout and confirm all cookies are cleared
- [ ] Verify redirect to home page after logout

### DIFF
--- a/components/frontend/src/components/navigation.tsx
+++ b/components/frontend/src/components/navigation.tsx
@@ -14,14 +14,9 @@ export function Navigation() {
   const router = useRouter();
 
   const handleLogout = () => {
-    // Clear all cookies
-    document.cookie.split(";").forEach((c) => {
-      document.cookie = c
-        .replace(/^ +/, "")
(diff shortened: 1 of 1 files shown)


components/frontend/src/components/navigation.tsx
```diff
@@ -13,6 +13,17 @@ export function Navigation() {
   // const segments = pathname?.split("/").filter(Boolean) || [];
   const router = useRouter();
 
+  const handleLogout = () => {
```
Gkrumbach07, line 16: it should be this

```tsx
const handleLogout = () => {
    // Redirect to oauth-proxy logout endpoint
    // This clears the OpenShift OAuth session and redirects back to login
    window.location.href = '/oauth/sign_out';
  };
```

Gkrumbach07 (changes requested):

PR #197, opened 2025-10-20
introduce a version set through environment variables

cooktheryan:


components/frontend/src/components/version-footer.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export function VersionFooter() {
+  const [version, setVersion] = useState<string | null>(null);
+
+  useEffect(() => {
+    fetch('/api/version')
+      .then((res) => res.json())
+      .then((data) => setVersion(data.version))
+      .catch((err) => console.error('Failed to fetch version:', err));
+  }, []);
```
Gkrumbach07: please create the api call in the service layer `components/frontend/src/services`

1. create a new file `version.ts` in `/api`
2. create a new file in queries called `use-version.ts`

follow existing patterns

Gkrumbach07 (changes requested):

PR #204, opened 2025-10-21
migrate runner from claude_code_sdk to claude_agent_sdk

MichaelClifford:


sallyom:
~I don't know why, but `/slash` commands don't work with claude_agent_sdk~  
EDIT: All good!

PR #227, opened 2025-10-30
Feature/docs cleanup v2

jeremyeder:
this was a huge cleanup. mostly removing placeholders but i also dropped the developer guide as not useful right now. i previewed this locally and its fine


jeremyeder:
still  not great but far better.

PR #243, opened 2025-11-04
Allow for the usage of Vertex

cooktheryan:


sallyom:
I will fix golangcilint errors in a follow-up - these are from a more strict config with golangci-lint v2 so will be many small nit fixes not related to this PR. Let's merge it!

PR #251, opened 2025-11-05
Add Gitlab support

natifridman:


RanPollak:
@natifridman  According to your latest update we are waiting for the UI changes as it's dependent to this contribution.
Only after the change we may have to solve some conflicts that may detect.

Update this info here so it's clear to all the relevant.

Gkrumbach07:
Tracked in Jira: https://issues.redhat.com/browse/RHOAIENG-39120 (same as PR #339)

dgutride:
this appears to be a duplicate of a newer one - moving to closed (please re-open if necessary)

PR #272, opened 2025-11-08
UI revamp fix backend

Gkrumbach07:


sallyom (approved):
There will be follow-up and many breaking changes :)

PR #273, opened 2025-11-09
update make local-dev target for base/decomposed kustomize templates

jeremyeder:
hey @Gkrumbach07 I am not sure whether you intended to shuffle paths when adding the e2e tests, but anyway

This PR
* updates the local-dev target to include the new `base` path
* consolidates operator build objects into build-configs.yaml


Gkrumbach07 (approved):

jeremyeder:
@bobbravo2 i dont think we need this anymore

PR #281, opened 2025-11-11
build fix frontend

sallyom:
quick fix


sallyom:
need to merge this to get the gh workflow fix (I could create a separate PR but have confirmed image build is fixed with this)

PR #291, opened 2025-11-12
add missing env vars to prod deploy

sallyom:
@cooktheryan I only updated the build-deploy script yesterday, missed this.


cooktheryan:
LGTM

PR #337, opened 2025-11-17
Feature/add codebase agent

jeremyeder:
draft for discussion. I think we would also want to add something in claude.md that points to this agent. we also need to centralize logic. Should ewe move out of claude.md "entirely", and into an agent(s)?

### DIFF
--- a/agents/amber.md
+++ b/agents/amber.md
@@ -159,6 +159,23 @@ flowchart TD
 - Request: "Use docker" → CLAUDE.md preference (podman) → Warn, ask confirmation
 - Request: "Add logging" → No conflicts → Implement with structured logging (constitution compliance)
 
+**Detailed Examples:**
+
+**Constitution Violations (Always Decline):**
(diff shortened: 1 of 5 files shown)


agents/amber.md
```diff
@@ -0,0 +1,541 @@
...
+
+**5. User Safety & Trust**
+- Act like you are on-call: responsive, reliable, and responsible
+- Always explain what you're doing and why before taking action
+- Provide rollback instructions for every change
+- Show your reasoning and confidence level explicitly
+- Ask permission before making potentially breaking changes
+- Make it easy to understand and reverse your actions
+- When uncertain, over-communicate rather than assume
+- Be nice but never be a sycophant—this is software engineering, and we want the CORRECT ANSWER regardless of feelings
```
jeremyeder, line 44: this is an important behavior change for claude

jeremyeder:
image

dgutride (approved):

PR #344, opened 2025-11-19
Fix: Add node affinity to prevent Multi-Attach PVC errors

### LINKED ISSUE #298: Bug: [restarted timed-out sessions fail to start due to multi-attach PVC error]
## 🐛 Bug Description

I'm seeing an issue where I started a session yesterday and I didn't stop it, so it ran until the time limit. I'd like to start back up today - I'm given a "continue session" option and I click that, I see my artifacts, so the temp content pod starts up, but the session doesn't successfully start due to multi-attach with PVC. We can fix this!
* when a session terminates due to time limit, don't necessarily fail (mine was in a failed state this morning)
* release any PVCs from timed-out sessions

jeremyeder:
Fixes #298

## Problem
AgenticSessions stuck in "Creating" phase due to temp-content and Job pods scheduled on different nodes with RWO PVCs.

## Solution
Add node affinity to ensure all containers schedule on the same node as the PVC attachment.

## Implementation
See `docs/implementation-prompts/fix-multi-attach-node-affinity.md` for complete cold-startable implementation instructions.

## Testing Needed
- [ ] Create new AgenticSession and verify no Multi-Attach errors
- [ ] Verify all containers scheduled on same node
- [ ] Test with cluster autoscaling enabled

---

**Note**: This is a draft PR with implementation guidance. Code changes still need to be implemented.

### DIFF
--- a/docs/implementation-prompts/fix-multi-attach-node-affinity.md
+++ b/docs/implementation-prompts/fix-multi-attach-node-affinity.md
@@ -1,0 +1,79 @@
+# Implementation: Fix Multi-Attach PVC Node Affinity
+
+## Problem
+AgenticSessions get stuck in "Creating" phase when the temp-content initContainer and the Job's main pods are scheduled on different nodes while using ReadWriteOnce (RWO) PVCs. This causes a Multi-Attach error because RWO PVCs can only be mounted on one node at a time.
+
+## Root Cause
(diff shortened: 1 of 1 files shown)


jeremyeder:
I haven't seen this in a long time. The fix is incorrect. I'm closing this and will get a proper fix if i see it again.

PR #220, opened 2025-10-29
Restrict OAuth scope to prevent console and kubectl access

jeremyeder:
Changed OAuth proxy scope from 'user:full' to 'user:info' to allow users to authenticate to vTeam via Google OAuth while blocking access to the OpenShift console and kubectl/oc CLI commands.

Changes:
- Updated frontend-deployment.yaml OAuth proxy scope
- Added security documentation to OPENSHIFT_OAUTH.md
- Created deployment guide in OAUTH_SCOPE_RESTRICTION.md

This provides app-only authentication without granting cluster privileges.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #346, opened 2025-11-19
Fix token exposure in frontend tool message display

jeremyeder:
Adds comprehensive token redaction to the frontend UI component that displays bash commands and tool inputs/outputs. This prevents sensitive tokens from being exposed in cleartext when viewing tool execution details.

Changes:
- Added redactSecrets() function to tool-message.tsx
- Applied redaction to tool input display (formatToolInput)
- Applied redaction to tool result content display
- Applied redaction to extracted result text (extractTextFromResultContent)

Redaction patterns:
- GitHub tokens (ghp_, ghs_, gho_, ghu_ prefixes)
- x-access-token: patterns in URLs
- OAuth tokens in URLs
- Basic auth credentials in URLs
- Authorization header values (Bearer tokens)
- Common API key patterns (sk-*, api_key, etc.)

This complements existing token redaction in:
- Backend: components/backend/server/server.go (query string redaction)
- Runner: components/runners/claude-code-runner/wrapper.py (command log redaction)

Fixes token exposure reported in bash command display.


cgwalters:
I'm sure there's some high level logic we could share with CI/CD systems all of which have this problem. 

A general best practice here I think will be ensuring that users can provide secret values in a structured way into their workflows via e.g. Kube secrets right?

Looks like https://github.com/tektoncd/pipeline/issues/3373 is a discussion around this.

jeremyeder:
Yes, we are going to need secrets management. Would Vault be alright?

cgwalters:
Hmm I would say as a baseline we should support providing Kubernetes secrets. Though this gets into a hugely important topic which is basically: To what degree do we expect the users of this platform to have an understanding of and/or control over the underlying Kubernetes environment?

I like the "workspace = namespace" model as a default, and it seems natural to ensure that users creating a workspace also can conveniently directly manipulate Kube native resources there, including secrets.

This topic also relates to https://github.com/ambient-code/platform/pull/364 in which I would like to start generalizing the *agent's* access to the underlying Kube cluster slightly, and make it also possible for the agent to do more.

jeremyeder:
You can use secrets all you want right now if they're secure enough for you.

Until you mentioned kubectl my intent was to disable user login via API and web. https://github.com/ambient-code/platform/pull/220#issue-3568038437

I did not want people getting the idea that this is a random OCP server to host stuff. If we decide to do something like #220 we can give your team's logins the right permissions to use kubectl + the OCP console. Thoughts?

cgwalters:
The thing that is actually quite important to me on the topic of secrets is that I want to make it really convenient and reliable to ensure that I'm providing to coding agents as precisely scoped credentials as possible.

And Github for example strongly encourages regularly rotating oauth tokens - tangent but I tried minting a token with an expiry of a year out of convenience, and it turns out that Github can allow organizations to deny access to such long-lived tokens, and in fact the CNCF does so by default.

Anyways I definitely think this project should encourage and help people to interact with such services that encourage (and even enforce) rotation.

I personally don't have much in the way of exfiltration concerns, so I can provide "read all repos" style tokens, but I *do* care about write privileges.

> I did not want people getting the idea that this is a random OCP server to host stuff.

Yeah of course, though exposing services/ingress is distinct from e.g. raw pods - I can definitely see use cases to making it easy for an agent to spin up parallel pods to investigate different approaches while working on a project and actually leveraging the cluster aspect to use multiple nodes.

Of course, some people who are developing e.g. kubernetes-native apps would find it quite convenient to be able to use services/ingress. Going to the inception level here - this project is obviously one!

---

Anyways sorry this is all kind of a digression, but the thing I would say we should do in this PR:

- Hiding based on well-known regexes various secret patterns: sure, though it'd be good to try to share code with other projects that want the same, and it may need to be configurable
- It would make sense to always redact anything that matches a secret value provisioned to the agent, regardless of whether it matches a regexp (and that's the connection with how we store secrets)

bobbravo2:
@jeremyeder @cgwalters Release engineering human here 👋🏾 . I wanted to see if we wanted to try to keep PR #346 on the https://github.com/ambient-code/platform/milestone/1 v0.0.14 train, or if we wanted to action the above and target [0.0.15](https://github.com/ambient-code/platform/milestone/2) as the next release? Thanks for your contributions!

Gkrumbach07:
Frontend redacting does nothing here. we should move this to backend to runner instead. Tessle stuff may need to be moved to seperate effort

PR #348, opened 2025-11-19
refactor: migrate claude-code-runner Dockerfile to use uv and uv.lock

bobbravo2:
## Changes

This PR refactors the claude-code-runner Dockerfile to use `uv` and `uv.lock` for faster, more reproducible builds.

### What Changed

- ✅ Added `uv` installation from official ghcr.io image
- ✅ Replaced `pip install` with `uv pip install` for runner-shell
- ✅ Use `uv sync --frozen --no-dev` for claude-runner dependencies
- ✅ Updated CMD to use `uv run --frozen` for reproducible execution
- ✅ Set `UV_SYSTEM_PYTHON=1` for system Python compatibility
- ✅ Copy `uv.lock` to ensure locked dependency versions

### Benefits

- 🚀 **Faster build times**: `uv` is significantly faster than `pip`
- 🔒 **Reproducible builds**: `uv.lock` ensures exact dependency versions across all builds
- 📦 **Better dependency resolution**: Superior conflict resolution
- 💾 **Smaller layers**: More efficient caching

### Testing

- [x] Dockerfile syntax validated
- [ ] Tested locally with podman build 
```bash
cd components/runners && podman build -f claude-code-runner/Dockerfile -t vteam-claude-runner:test .
```

- [ ] Build test needed in CI

### Related

- Part of ongoing effort to standardize on `uv` for Python package management
- Supersedes #347 (closed in favor of this clean branch)


bobbravo2:
## Addressed PR Feedback

Thank you for the comprehensive review! I've addressed all the recommendations:

### ✅ Completed

**Major Issues:**
1. **Build Context Documentation** - Added comment at top of Dockerfile documenting the expected build context
2. **CI Build Context Validation** - Verified GitHub Actions workflow (lines 64-68) correctly uses:
   - `context: ./components/runners`
   - `dockerfile: ./components/runners/claude-code-runner/Dockerfile`

**Minor Issues:**
3. **Layer Optimization** - Separated dependency installation from code copy:
   - Dependencies copied first (line 35)
   - Dependencies installed (line 39-43)
   - Application code copied last (line 46)
   - This improves caching - when only wrapper.py changes, dependency layer remains cached

4. **Command Documentation** - Added comprehensive comment explaining why `uv run` is necessary:
   - Dependencies are isolated in `/app/claude-runner/.venv`
   - `uv run --frozen` activates the venv and uses locked dependencies
   - Verified this is required (tested both approaches)

5. **Inline Comments** - Added detailed comments throughout explaining:
   - `UV_SYSTEM_PYTHON=1` purpose
   - Build context requirements
   - `uv sync` vs `uv pip install` rationale
   - `--frozen` flag importance
   - OpenShift compatibility permissions

### 🧪 Testing

✅ Built successfully with podman:
```bash
cd components/runners
podman build -f claude-code-runner/Dockerfile -t vteam-claude-runner:test .
```

✅ Verified all key dependencies are available:
- anthropic 0.68.0
- claude-agent-sdk 0.1.4
- requests 2.32.5
- pyjwt 2.10.1

✅ Confirmed layer caching optimization works - most layers cached on rebuild

Ready for merge! 🚀

bobbravo2:
## Response to Latest Review Comments

Thank you for the detailed review! Let me address the concerns:

### Blocker Issues

**1. ✅ Missing uv.lock file validation**
- **Status**: uv.lock exists and is committed to git (190KB file)
- **Verification**:
  ```bash
  $ ls -lh components/runners/claude-code-runner/uv.lock
  -rw-r--r--@ 1 bgregor  staff   190K Nov 14 18:18 uv.lock
  
  $ git ls-files components/runners/claude-code-runner/uv.lock
  components/runners/claude-code-runner/uv.lock
  ```
- **CI Confirmation**: The build passed (6m42s), which validates the lock file works correctly

**2. 🤔 Inconsistent package management**
- **Status**: This is intentional design, but I understand the concern
- **Rationale**:
  - `runner-shell` is a **shared base framework** (like a system library)
  - `claude-runner` is the **application** with specific locked dependencies
  - Pattern: Base framework installed globally, application dependencies isolated in venv
- **Trade-off**: You're right that runner-shell isn't locked in this PR. However:
  - This is a separate concern from migrating claude-runner to uv
  - runner-shell could be migrated to use uv.lock in a separate PR
  - For now, maintaining the existing pattern for runner-shell minimizes risk
- **Alternative**: I could create runner-shell/uv.lock and use `uv sync` there too, but that's a larger change

### Critical Issues

**3. ✅ UV_SYSTEM_PYTHON=1 bypasses isolation**
- **Rationale**: We're using the Python from the base image (`python:3.11-slim`)
- **Why needed**: Without this flag, uv would try to manage its own Python installations
- **Security**: The venv still provides isolation - dependencies are in `/app/claude-runner/.venv`
- **Documentation**: Already added in the Dockerfile (line 25)

### CI Status

✅ **All checks passing**:
- claude-code-runner build: pass (6m42s)
- End-to-End Tests: pass (3m41s)
- All other builds: pass

The build system validates that:
- uv.lock is present and valid
- Dependencies resolve correctly
- The image builds successfully for linux/amd64 and linux/arm64

### Recommendation

The current implementation is production-ready:
- ✅ uv.lock is committed and validated by CI
- ✅ UV_SYSTEM_PYTHON=1 is documented and necessary
- ⚠️ runner-shell could use uv.lock (enhancement for future PR)

**Suggested path forward**:
1. Merge this PR (claude-runner migration is complete and tested)
2. Open follow-up issue: "Migrate runner-shell to use uv.lock" for consistency

What do you think? 🤔

bobbravo2:
## Fixed: Missing CLAUDE.md Reference

Good catch! You were absolutely right about this issue.

### The Problem
The `pyproject.toml` referenced:
```toml
readme = "CLAUDE.md"
```

But `CLAUDE.md` doesn't exist in the `claude-code-runner` directory. When the Dockerfile was optimized to copy only specific files (pyproject.toml, uv.lock, wrapper.py), it would fail to find the readme during `uv sync`.

**Why my builds succeeded:** uv appears to handle missing readme files gracefully (it's optional metadata), so the build didn't fail. However, this is still a latent bug that should be fixed.

### The Fix
✅ Removed the `readme = "CLAUDE.md"` line from pyproject.toml  
✅ Regenerated `uv.lock` to reflect the updated metadata  
✅ Verified build and runtime still work correctly

### Testing
```bash
# Build test
cd components/runners
podman build -f claude-code-runner/Dockerfile -t test .
# ✅ SUCCESS

# Runtime test  
podman run --rm test bash -c "cd /app/claude-runner && uv run --frozen python -c 'import anthropic; import claude_agent_sdk; print(\"OK\")'"
# ✅ All dependencies working
```

Thank you for catching this! 🙏

jeremyeder:
@sallyom @bobbravo2 do we still need this?

bobbravo2:
> @sallyom @bobbravo2 do we still need this?

I think this is obsolete! Closed.

PR #364, opened 2025-11-25
docs: add ADR-0006 for agent injection architecture

cgwalters:
Add architectural decision record documenting how to support user-provided container images with agent code injection. This addresses the requirement to let users run agentic sessions in their own container images with their own tools without forcing them to extend our base image.


jeremyeder:
@cgwalters hey, here is what I was talking about bring-your-own: https://github.com/ambient-code/workflows/tree/main/workflows/template-workflow#directory-structure

All that stuff gets cloned into your session PVC to be made available to claude and agents.

I am using Workflows as the top level thing that teams use to "onboard" to the system.

Workflows don't include a way to customize the runner image, either by having a custom image or customizing the system's base image. So...

@MichaelClifford any concerns with this?

cgwalters:
FTR I'm keeping this draft for now, I'm still refining this and playing with the experimental implementation.

cgwalters:
Should we try to separate a pull request adding an ADR from one adding an implementation? (Kind of like RFC vs code?)

cgwalters:
OK I completely reworked this as I decided it's just way simpler to skip things like custom MCP servers and basically give the agent sugar on top of `kubectl` as a way to create pods for its work.

I also dropped out the PoC implementation work I had; feels cleaner to keep that as a separate PR. Going to have the agent try updating the PoC to match this latest implementation proposal to see how it works/feels.

cgwalters:
I'm continuing to iterate on this and experiment with a PoC implementation, I'll post an update soon

components/manifests/overlays/minikube/namespace-patch.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
```
cgwalters: Oops sorry, some minikube-related stuff leaked into this PR too beacuse I'd been developing on top of https://github.com/ambient-code/platform/pull/367 first. Will fix.

components/runners/claude-code-runner/patches/README.md
```diff
@@ -0,0 +1,149 @@
+# Claude Agent SDK Patches
+
+This directory contains patches applied to the `claude-agent-sdk` package to ensure compliance with ADR-0006 (Agent Injection Architecture).
```
cgwalters, line 3: This was good for PoC but now we need to actually try upstreaming these changes without patching like this.

components/scripts/minikube/deploy.sh
```diff
@@ -0,0 +1,233 @@
+#!/bin/bash
+# Minikube deployment script for Ambient Code Platform
```
cgwalters: Also remove this

cgwalters:
OK, there's now conflicts with https://github.com/ambient-code/platform/pull/367 on git main (which I was using to develop this). I may try testing it this out instead on a full OCP (or maybe https://github.com/minc-org/minc instead of minikube), but anyways this is probably far enough along that folks can comment on the overall architecture idea.

Does anyone think it might make sense to just always have a complete "sub-namespace" per workspace where the agent can run arbitrary pods by default say? Or maybe we go all the way to requiring https://github.com/projectcapsule/capsule (feels a bit like overkill by default, but we should certainly make it easy to integrate with that, as well as being easily able to provision completely distinct cloud resources that are lifecycle-bound; lots of precedent in that space).

cgwalters:
Update on this: I'm still interested in this and working on it, but right now I've switched to doing a spike on https://github.com/loft-sh/devpod/ which has a lot of nice aspects to it and looking at what AI tooling augmenting that might look like.

jeremyeder:
hey @cgwalters unfortunate this didn't work out. maybe we can find an alignment in the future.

cgwalters:
> hey @cgwalters unfortunate this didn't work out. maybe we can find an alignment in the future.

I'm still interested in that! I think the simplest path in a nutshell is:

- Ambient gains support for .devcontainer (like devaipod but also like devpod and the official devcontainer CLI)
- Ambient ships a devcontainer ["feature"](https://containers.dev/implementors/features/) that can be used to install its agent glue. (Or of course in my ideal world, cuts over to opencode and then just ask people to install opencode in their devcontainer, which is what devaipod does)

That should probably be it.

PR #365, opened 2025-11-25
add mcp-atlassian to runner for jira use

MichaelClifford:


components/runners/claude-code-runner/wrapper.py
```diff
@@ -2024,72 +2024,30 @@ def _filter_mcp_servers(self, servers: dict) -> dict:
         return allowed_servers
 
     def _load_mcp_config(self, cwd_path: str) -> dict | None:
-        """Load MCP server configuration from .mcp.json file in the workspace.
+        """Load MCP server configuration from the vTeam runner's .mcp.json file.
```
aegeiger: Nitpick: didn't we say we want to stop using this term?

aegeiger (approved):
LGTM, for what it worth
IMHO this is two different changes, add mcp-atlassian and remove support for users' MCPs. I think the later is a good idea but it kind-of hides in this other change. I don't remember if we ever said we support users' MCPs but if we did we need to communicate this change clearly.

components/runners/claude-code-runner/wrapper.py
MichaelClifford: Good catch! changing!

MichaelClifford:
@aegeiger I completely agree its two changes. But given the high priority of adding this MCP server, and it relying on the separate change of moving where MCP's are defined I hope its ok to make an exception here.

aegeiger:
> @aegeiger I completely agree its two changes. But given the high priority of adding this MCP server, and it relying on the separate change of moving where MCP's are defined I hope its ok to make an exception here.

I humbly agree.

sallyom (approved):
LGTM - we'll consider how users can/should BYO MCPs in a follow-up.

PR #366, opened 2025-11-25
docs: adopt AGENTS.md standard for agent instructions

cgwalters:
Rename CLAUDE.md to AGENTS.md to align with the community AGENTS.md convention (https://agents.md/). This standardized format makes the repository's agent instructions more discoverable and consistent with other projects using AI coding assistants.

My real motivation here is I want to support other agentic systems and not having this project hardcoded to Claude Code, which is proprietary software.

Assisted-by: Claude Code (Sonnet 4.5)


cgwalters:
OK yeah, man there's a ton of hardcoded references. I guess we may as well try to go a little bit farther here, but if folks feel differently I'm happy to just do the symlink effectively.

cgwalters:
> Following AGENTS.md Standards: ✅ Yes - ironic but appropriate given the PR content

haha

jeremyeder:
symlink should be fine

natifridman:
I think symlink should be enough. I don't think there is need to change the reference in the other files

cgwalters:
I rebased :surfer: - @jeremyeder mind reviewing?

jeremyeder:
fixed via #429

cgwalters:
There's a subtle but important difference between the two: this PR was trying to start the ball rolling on not having everything be about Claude Code in this repository.

But we can debate that here https://github.com/ambient-code/platform/issues/431

And if that issue is accepted, then I think it would make sense to have CLAUDE.md be the symlink to AGENTS.md, not the other way around.

PR #378, opened 2025-11-27
fix: Add error handling to git operations in Claude Code runner

aegeiger:
Wrap git clone, fetch, checkout, and reset operations in try-catch blocks to prevent session failures when git operations encounter errors. Failed git operations now log warnings and allow sessions to continue gracefully.

This improves resilience when:
- Repository URLs are invalid or unreachable
- Network issues occur during clone/fetch
- Branch references don't exist
- Authentication fails for specific repos in multi-repo sessions

### DIFF
--- a/components/runners/claude-code-runner/tests/test_git_error_handling_simple.py
+++ b/components/runners/claude-code-runner/tests/test_git_error_handling_simple.py
@@ -1,0 +1,196 @@
+"""
+Simplified tests for git operation error handling in wrapper.py
+
+These tests verify the core error handling improvements:
+1. Specific exception types are caught (RuntimeError, OSError not broad Exception)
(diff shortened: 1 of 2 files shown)


jeremyeder:
hey @aegeiger can you show me what this looks like when it happens? How often? How much time does it save when it catches something?

aegeiger:
Initially it was meant to address a concern of @sallyom that it's hard to debug when a failure to clone happens, for instance if the user tries to clone a repo from a private gitlab we don't have access to.  But more broadly, I think her point was right, that a failure to clone shouldn't result in a failed session.

jeremyeder:
i havent seen or heard folks hit this, lets table this for now and re-open if it gets reported again.

Issue #380, opened 2025-11-28
[Amber] sessions timeout and cant resume


jeremyeder:
### Issue Description

I'm having an issue with this application where if I leave an active session idle for a while (unsure exactly how long, but perhaps less than 1 hr), it just gets stuck. I can send chat messages but claude never replies. Hitting stop/resume on the chat doesn't do anything.

Once I click resume, the pod goes from creating straight to failed.

So 2 problems:

1. Why do idle sessions just silently die? Do we need a keepalive for running sessions?
2. When I click stop on a dead session, it tries to resume, and the session goes from creating straight to failed.

### Fix Type

Code Formatting

### Confirmation

- [x] I understand this will create an automated PR
- [x] The changes are low-risk and reversible
- [x] All tests should continue to pass after fixes

PR #388, opened 2025-11-28
feat(automation): add retry logic for GitHub API label addition in Amber workflow

jeremyeder:
## Summary
Add exponential backoff retry pattern to harden PR label addition against transient GitHub API failures.

## Problem
Amber workflow occasionally fails with `RequestError [HttpError]: Unexpected end of JSON input, status: 500` when adding labels to PRs. This is a transient GitHub API error that should be retried.

## Solution
- Add `retryWithBackoff()` helper function with exponential backoff
- 3 retry attempts with 1s, 2s, 4s delays between attempts
- Only retries on retriable errors (HTTP 5xx, JSON parse errors)
- Fails fast on client errors (4xx) since retrying won't help

## Test Plan
- [x] Workflow logic reviewed
- [ ] Test with intentional API failure (manual testing required)
- [ ] Monitor next Amber automation run for successful label addition

## Changes
- Modified `.github/workflows/amber-issue-handler.yml`:
  - Added retry helper function (lines 324-342)
  - Wrapped label addition with retry logic (lines 379-387)

## References
- Related to successful PR #387 (labels failed due to transient API error)
- Addresses issue #380 follow-up (workflow hardening)

🤖 Generated with Claude Code (https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/amber-issue-handler.yml
+++ b/.github/workflows/amber-issue-handler.yml
@@ -322,23 +322,29 @@ jobs:
             const repository = process.env.GITHUB_REPOSITORY;
 
             // Helper function for retrying API calls with exponential backoff
+            // Retries on: 5xx errors, network errors (no status), JSON parse errors
             async function retryWithBackoff(fn, maxRetries = 3, initialDelay = 1000) {
               for (let i = 0; i < maxRetries; i++) {
                 try {
                   return await fn();
                 } catch (error) {
                   const isLastAttempt = i === maxRetries - 1;
-                  const isRetriable = error.status >= 500 || error.message.includes('JSON');
+                  // Retry on: network errors (undefined status), 5xx errors, or specific error patterns
+                  const isRetriable = !error.status || error.status >= 500;
 
                   if (isLastAttempt || !isRetriable) {
                     throw error;
                   }
 
                   const delay = initialDelay * Math.pow(2, i);
-                  console.log(`Attempt ${i + 1} failed, retrying in ${delay}ms...`);
+                  const errorMsg = error.message || 'Unknown error';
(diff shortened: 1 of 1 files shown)


jeremyeder:
## Review Feedback Addressed ✅

I've addressed the high-priority feedback from the automated code review:

### 1. Enhanced Retry Detection for Network Errors
**Before:**
```javascript
const isRetriable = error.status >= 500 || error.message.includes('JSON');
```

**After:**
```javascript
// Retry on: network errors (undefined status), 5xx errors, or specific error patterns
const isRetriable = \!error.status || error.status >= 500;
```

**Why:** Network errors (timeouts, connection resets, DNS failures) don't have a `.status` property. The new logic assumes all errors are retriable except 4xx client errors, which is more robust.

### 2. Added Error Details to Logs
**Before:**
```javascript
console.log(`Attempt ${i + 1} failed, retrying in ${delay}ms...`);
```

**After:**
```javascript
const errorMsg = error.message || 'Unknown error';
const errorStatus = error.status || 'network error';
console.log(`Attempt ${i + 1} failed (${errorStatus}: ${errorMsg}), retrying in ${delay}ms...`);
```

**Why:** Debugging is much easier when you can see exactly what error occurred on each retry attempt.

### 3. Added Explicit Final Throw
**After the loop:**
```javascript
// Defensive: Should never reach here due to throw in loop, but explicit for clarity
throw new Error('retryWithBackoff: max retries exceeded');
```

**Why:** Makes the function's contract crystal clear and prevents unreachable code paths.

---

These changes make the retry logic more robust against network failures while improving debuggability. Ready for re-review\! 🚀

PR #418, opened 2025-12-04
fix: Only validate ambient-runner-secrets when Vertex AI is disabled

sallyom:
Fixes a regression where ambient-runner-secrets validation occurs unconditionally, causing session creation to fail even when CLAUDE_CODE_USE_VERTEX=1.

This fix wraps the runner secret validation in a !vertexEnabled check, since the secret is only needed when Vertex AI is disabled. When Vertex is enabled, the ambient-vertex secret is used instead.

Changes:
- Wrap runner secret validation in 'if !vertexEnabled' block
- Add descriptive log messages for both Vertex enabled/disabled cases
- Maintain error handling and status conditions for non-Vertex case


sallyom:
testing now

sallyom:
@Gkrumbach07 not sure about the Amber review, but this fixes the issue with Vertex-enabled clusters.

PR #429, opened 2025-12-04
feat: add AGENTS.md symlink support with comprehensive testing

jeremyeder:
## Summary

Add support for Cursor and other AI tools that use `AGENTS.md` by creating a symlink to `CLAUDE.md`. This provides zero-maintenance dual-tool support with comprehensive validation.

## Changes

- ✅ Create `AGENTS.md` → `CLAUDE.md` symlink for Cursor compatibility
- ✅ Add minimal inline symlink validation to e2e workflow (fails fast)
- ✅ Add comprehensive standalone test script (10 checks)
- ✅ Add dedicated GitHub Actions workflow for symlink validation
- ✅ Add `tests/README.md` documenting the testing strategy
- ✅ Update `.gitignore` to exclude `.cursor/` and `.tessl/` directories

## Testing Strategy

Three-tier validation approach:

1. **E2E inline check** - Fast validation on every e2e run (~1 second)
   - Runs after checkout, before builds
   - Fails fast to save CI time
   - 4 essential checks (exists, is symlink, target, content match)

2. **Dedicated workflow** - Runs when `AGENTS.md` or `CLAUDE.md` change
   - Comprehensive validation including git tracking
   - Platform compatibility checks

3. **Local test script** - `tests/test_agents_md_symlink.sh`
   - 10 detailed checks with informative error messages
   - Validates project context and key documentation sections

## Why a Symlink?

**Problem**: Claude Code uses `CLAUDE.md`, Cursor uses `AGENTS.md`

**Solution**: Symlink eliminates duplication and maintenance overhead

**Benefits**:
- Zero maintenance (single source of truth)
- No sync issues between files
- Git tracks symlinks correctly across platforms
- Works on macOS, Linux, WSL, and Windows (Git for Windows)

## Test Results

```
✅ All tests passed! AGENTS.md symlink is working correctly.
   - Symlink: AGENTS.md -> CLAUDE.md
   - Content size: 36394 bytes
   - Cursor and other AI tools can use AGENTS.md
```

## Test Plan

- [x] Local test script passes all 10 checks
- [ ] E2E workflow runs successfully with new inline validation
- [ ] Dedicated symlink workflow validates correctly
- [ ] Cursor can read `AGENTS.md` successfully (manual verification)

### DIFF
--- a/.github/workflows/test-agents-md-symlink.yml
+++ b/.github/workflows/test-agents-md-symlink.yml
@@ -1,0 +1,64 @@
+name: Test AGENTS.md Symlink
+
+on:
+  pull_request:
+    paths:
+      - 'AGENTS.md'
+      - 'CLAUDE.md'
+      - 'tests/test_agents_md_symlink.sh'
+      - '.github/workflows/test-agents-md-symlink.yml'
+  push:
+    branches:
+      - main
+    paths:
+      - 'AGENTS.md'
+      - 'CLAUDE.md'
+      - 'tests/test_agents_md_symlink.sh'
+
+jobs:
+  test-symlink:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v5
+
+    - name: Validate AGENTS.md symlink
+      run: |
+        chmod +x tests/test_agents_md_symlink.sh
+        ./tests/test_agents_md_symlink.sh
+
+    - name: Test cross-platform compatibility
+      run: |
+        echo "Testing symlink on Linux..."
+
+        # Verify symlink properties
+        ls -la AGENTS.md
+
+        # Verify git stores it correctly
+        git ls-files -s AGENTS.md
+
+        # Verify content is accessible
+        head -n 5 AGENTS.md
+
+        echo "✅ Symlink works correctly on Linux"
+
+    - name: Validate git tracking
+      run: |
+        # Check that AGENTS.md is in git index
+        if git ls-files --error-unmatch AGENTS.md > /dev/null 2>&1; then
+          echo "✅ AGENTS.md is tracked by git"
+        else
+          echo "❌ ERROR: AGENTS.md is not tracked by git"
+          exit 1
+        fi
+
+        # Check that it's stored as a symlink in git
+        MODE=$(git ls-files -s AGENTS.md | awk '{print $1}')
+        if [ "$MODE" = "120000" ]; then
+          echo "✅ AGENTS.md is stored as symlink (mode 120000)"
+        else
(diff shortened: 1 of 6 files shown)


.github/workflows/test-agents-md-symlink.yml
```diff
@@ -0,0 +1,64 @@
...
+
+        # Verify git stores it correctly
+        git ls-files -s AGENTS.md
+
+        # Verify content is accessible
+        head -n 5 AGENTS.md
+
+        echo "✅ Symlink works correctly on Linux"
+
+    - name: Validate git tracking
```
cgwalters, line 47: er...a whole github actions workflow to verify a file is in the git repository?

That feels like massive overkill?

tests/test_agents_md_symlink.sh
```diff
@@ -0,0 +1,121 @@
+#!/bin/bash
+
+# Test that AGENTS.md symlink works for Cursor and other AI tools
+# This validates that:
+# 1. AGENTS.md exists and is a valid symlink
+# 2. It points to CLAUDE.md
+# 3. Content is readable and identical to CLAUDE.md
+# 4. File is tracked by git correctly
```
cgwalters, line 8: This testing feels pretty excessive for something so simple. What I think would be enough testing and more useful would be to verify some agent other than claude (like goose or whatever) running on this repo has successfully read in the content.

cgwalters:
To be clear is this intended to replace https://github.com/ambient-code/platform/pull/366 or is it a smaller preparatory step and we could still merge that PR or a variant of it that is actually aiming to not have everything be about Claude?

tests/README.md
```diff
@@ -0,0 +1,66 @@
+# Tests
+
+This directory contains project-level tests for the Ambient Code Platform.
+
+## Test Structure
+
+- `integration/` - Integration tests requiring real Kubernetes clusters
+- `test_agents_md_symlink.sh` - Validates AGENTS.md symlink for Cursor compatibility
```
cgwalters, line 8: How about generalizing this to something like `structure/` which is tests that statically verify the git repository state (without building anything). Could also be written in Go perhaps and not shell script.

tests/README.md
```diff
@@ -0,0 +1,66 @@
...
+### What It Tests
+
+1. ✅ AGENTS.md exists and is a valid symlink
+2. ✅ Symlink points to CLAUDE.md
+3. ✅ Content is readable through symlink
+4. ✅ Content matches CLAUDE.md exactly
+5. ✅ File is tracked by git correctly (mode 120000)
+6. ✅ Contains expected project context
+7. ✅ All required documentation sections exist
+8. ✅ File size is reasonable
```
cgwalters, line 23: And all of this stuff is already described in the test itself - what value is there in duplicating it? It can only get out of date.

PR #430, opened 2025-12-04
adding internal workflows doc

nsingla:
As a new contributor, I was struggling to completely understand how different components of the project interact with each other. So I created these worflow diagrams for my sanity and thought these might be useful to other new contributors like me as well.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -26,6 +26,8 @@ The platform includes **Amber**, a background agent that automates common develo
 
 ### Core Architecture
 
+For detailed visual workflows, see [Architecture Workflows](docs/architecture-workflows.md).
+
 The system follows a Kubernetes-native pattern with Custom Resources, Operators, and Job execution:
 
 1. **Frontend** (NextJS + Shadcn): Web UI for session management and monitoring
--- a/README.md
+++ b/README.md
@@ -550,6 +550,7 @@ See [e2e/README.md](e2e/README.md) for detailed documentation, troubleshooting, 
 ### Deployment & Configuration
 - **Deployment Guide**: [docs/OPENSHIFT_DEPLOY.md](docs/OPENSHIFT_DEPLOY.md)
 - **OAuth Setup**: [docs/OPENSHIFT_OAUTH.md](docs/OPENSHIFT_OAUTH.md)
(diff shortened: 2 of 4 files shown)


cgwalters:
The reviews mention mermaid...but I don't see the mermaid sources here? I don't think we even need pre-rendered png files, just have the mermaid in the markdown and it will get rendered when viewing by many clients.

Also crucially, AI models can natively edit Mermaid to keep it up to date as the architecture changes, which is a lot harder with pre-rendered PNG files.

nsingla:
> The reviews mention mermaid...but I don't see the mermaid sources here? I don't think we even need pre-rendered png files, just have the mermaid in the markdown and it will get rendered when viewing by many clients.
> 
> Also crucially, AI models can natively edit Mermaid to keep it up to date as the architecture changes, which is a lot harder with pre-rendered PNG files.

The mermaid code block is collapsed by default if you are in the view mode but its there https://github.com/ambient-code/platform/pull/430/files#diff-982cfd3f63cdcf05c7c1f75df2b6e5ba9f0b00f6aaa8b4908ce07258843aad8eR16

I added pre-rendered images so that one can read it directly on github as well. And yes, you are right,  AI models can natively edit mermaid block, but those can also auto generate corresponding png as well to keep it upto date which is what I did to implement some of the PR suggestion by Claude code review workflow

Let me know if you still feel strongly about removing the pre-rendered pngs

cgwalters:
> I added pre-rendered images so that one can read it directly on github as well. 

Github's default markdown renderer renders mermaid. There's already mermaid in this repository, so you can see an example here: https://github.com/ambient-code/platform/blob/2be79d29fdde64619a3c5888036a7ab389dbb1a7/docs/diagrams/amber-workflow.md

I don't think there's any advantage to pre-rendered PNGs.

nsingla:
> > I added pre-rendered images so that one can read it directly on github as well.
> 
> Github's default markdown renderer renders mermaid. There's already mermaid in this repository, so you can see an example here: https://github.com/ambient-code/platform/blob/2be79d29fdde64619a3c5888036a7ab389dbb1a7/docs/diagrams/amber-workflow.md
> 
> I don't think there's any advantage to pre-rendered PNGs.

oh ok, I thought it doesn;t. Let me remove the prerendered images then, thank you so much :bow:

bobbravo2 (approved):

jeremyeder:
point taken - i've created an issue to track documenting architecture 👍

PR #453, opened 2025-12-09
fix(amber): Push to existing PR instead of creating duplicates

natifridman:
When an issue already has an open PR, Amber now:
- Detects the existing PR via gh pr list search
- Checks out the existing branch instead of creating a new one
- Pushes additional commits to the existing PR
- Adds comments to both PR and issue about the update

This prevents duplicate PRs like #450, #442, #441, #438.

### DIFF
--- a/.github/workflows/amber-issue-handler.yml
+++ b/.github/workflows/amber-issue-handler.yml
@@ -1,3 +1,22 @@
+# Amber Issue-to-PR Handler
+#
+# This workflow automates issue resolution via the Amber background agent.
+#
+# TRIGGERS:
+# - Issue labeled with: amber:auto-fix, amber:refactor, amber:test-coverage
+# - Issue comment containing: /amber execute or @amber
+#
+# BEHAVIOR:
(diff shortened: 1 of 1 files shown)


.github/workflows/amber-issue-handler.yml
```diff
@@ -186,12 +205,29 @@ jobs:
...
+          # Validate issue number is numeric to prevent injection
+          if ! [[ "$ISSUE_NUMBER" =~ ^[0-9]+$ ]]; then
+            echo "Error: Invalid issue number format"
+            exit 1
+          fi
+
+          # Check if there's already an open PR for this issue using stricter matching
+          # Search for PRs that reference this issue and filter by body containing exact "Closes #N" pattern
+          EXISTING_PR=$(gh pr list --state open --json number,headRefName,body --jq \
+            ".[] | select(.body | test(\"Closes #${ISSUE_NUMBER}($|[^0-9])\")) | {number, headRefName}" \
```
jeremyeder, line 217: should be enough to break this bot loop

PR #465, opened 2025-12-15
Fix Makefile pod name

andybraren:
I notice that a bunch of our recent PRs are failing one of their checks. According to Claude's analysis:

> Summary
Problem: The test-local-dev-simulation job was failing because the Makefile was looking for pods with label app=backend, but the actual backend deployment uses app=backend-api.
Evidence: The troubleshooting output showed all pods were actually ready (1/1 Running, Ready: True), but the test was using the wrong label selector and couldn't find them.
Fix: Updated the Makefile in two locations:
Line 359 - local-test-quick target: Changed app=backend → app=backend-api
Line 606 - _auto-port-forward target: Changed app=backend → app=backend-api
This aligns the Makefile with the actual label used in components/manifests/base/backend-deployment.yaml. The test should now pass in CI. ✅

If this sounds right to folks then this PR should fix it. Let's see if it passes now.

EDIT: it does. 🎉


bobbravo2:
This should be obsolete now: https://github.com/ambient-code/platform/pull/460 has all checks passing. 

@andybraren please 💀 keep alerting us if we see regressions or checks starting to go 🔴. Thank you!

PR #468, opened 2025-12-16
Automatically refresh Artifacts panel

andybraren:
This PR triggers an automatic refresh of the Artifacts panel after every completed tool call message and completed session so that users don't need to click the Refresh button manually.

It also adds a badge to the panel's header whenever artifacts exist, which should make it easier to spot that artifacts have been created when the panel is collapsed (like it is by default).

cc @Daniel-Warner-X 

## Demo

https://github.com/user-attachments/assets/f430bfeb-660b-4d20-a0ea-b12c6b3d1156


Gkrumbach07:
id like this to eventually be replaced by some kind of streaming / shared state. but for now looks good

PR #498, opened 2026-01-08
Display git currently checked-out branch in context and file explorer modals

sallyom:
* Show the branch name as a badge next to each repository in the Context accordion and in the File Explorer. This helps users quickly identify which branch each repository is using.
* Properly handle multiple branches of same repositories
* This PR is included here: https://github.com/ambient-code/platform/pull/502

### DIFF
--- a/components/backend/handlers/helpers.go
+++ b/components/backend/handlers/helpers.go
@@ -48,6 +48,14 @@ func RetryWithBackoff(maxRetries int, initialDelay, maxDelay time.Duration, oper
 	return fmt.Errorf("operation failed after %d retries: %w", maxRetries, lastErr)
 }
 
+// ComputeAutoBranch generates the auto-branch name from a session name
+// This is the single source of truth for auto-branch naming in the backend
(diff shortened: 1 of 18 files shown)


sallyom:
The git repo context needs more of an overhaul than this PR- the platform needs to be aware of remotes and branches in general. This PR offers a slight improvement, but doesn't help with the platform's overall awareness of branches and remotes. 
I'm seeing an edge case where if I add a 2nd branch as context, the platform doesn't know what to do with it - the frontend happily reports it's been cloned - but the actual pod does not include the new branch:

```
(app-root) sh-5.1$ git branch
* git-context-show-branch
  main
```

Screenshot 2026-01-08 at 11 53 49 AM

sallyom:
@Gkrumbach07 Here's the banners with the generated branch name (`ambient/timestamp`)  
Screenshot 2026-01-12 at 5 20 57 PM

Gkrumbach07:
Why use timestamp, can't you use the session id so it's easier to trace back

PR #502, opened 2026-01-09
fix: Fresh client per turn + custom restart tool for Claude SDK

Gkrumbach07:
## Summary

Fixes hanging behaviour on second message by ensuring fresh Claude SDK client per message turn, plus adds a custom MCP tool for Claude to restart itself.

### Fresh Client Per Turn Pattern
- Each message creates a new `ClaudeSDKClient` (connect → query → receive → disconnect)
- Using `continue_conversation=True` maintains state from disk (`.claude/` directory)
- Prevents the SDK blocking issue where `receive_response()` hangs indefinitely on subsequent calls

### Race Condition Fix
- Moved `current_message_id` from instance variable to local variable in `_run_claude_agent_sdk`
- Prevents cross-contamination between concurrent runs

### Custom Restart MCP Tool (NEW)
- Added `mcp__session__restart_session` tool that Claude can call to request a session restart
- Uses `@tool` decorator and `create_sdk_mcp_server()` from the Claude Agent SDK
- When called, sets a flag and emits `session_restart_requested` event after run completes
- Allows Claude to recover from broken states or request a fresh connection

### Feature Branch on Runtime Clone
- `clone_repo_at_runtime` creates `ambient/<session-id>` branches when repos are added dynamically
- Removed from `hydrate.sh` startup (repos stay on original branch at init)

### Idempotent Repo Addition
- `/add-repo` endpoint now returns `was_newly_cloned` flag
- Prevents duplicate notifications when both backend and operator call the endpoint

## Trade-offs

**Performance:**
- ❌ ~1-2 second overhead per message from subprocess initialisation
- ✅ System actually works without hanging

**Correctness:**
- ✅ Conversation history maintained via disk state
- ✅ No blocking or freezing
- ✅ Reliable multi-turn conversations
- ✅ Claude can self-recover via restart tool

## Testing

Tested with multiple consecutive messages - no hanging, proper conversation flow maintained.

### DIFF
--- a/components/runners/claude-code-runner/adapter.py
+++ b/components/runners/claude-code-runner/adapter.py
@@ -70,9 +70,9 @@ class ClaudeCodeAdapter:
         self._skip_resume_on_restart = False
         self._turn_count = 0
 
-        # AG-UI streaming state
-        self._current_message_id: Optional[str] = None
-        self._current_tool_id: Optional[str] = None
+        # AG-UI streaming state (per-run, not instance state)
+        # NOTE: _current_message_id and _current_tool_id are now local variables
+        # in _run_claude_agent_sdk to avoid race conditions with concurrent runs
         self._current_run_id: Optional[str] = None
         self._current_thread_id: Optional[str] = None
         
@@ -289,6 +289,9 @@ class ClaudeCodeAdapter:
             thread_id: AG-UI thread identifier
             run_id: AG-UI run identifier
         """
+        # Per-run state - NOT instance variables to avoid race conditions with concurrent runs
+        current_message_id: Optional[str] = None
+        
         logger.info(f"_run_claude_agent_sdk called with prompt length={len(prompt)}, will create fresh client")
         try:
             # Check for authentication method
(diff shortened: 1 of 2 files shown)


components/runners/claude-code-runner/adapter.py
```diff
@@ -1300,25 +1340,34 @@ def _build_workspace_context_prompt(self, repos_cfg, workflow_name, artifacts_pa
 
         # Repositories
         if repos_cfg:
+            session_id = os.getenv('AGENTIC_SESSION_NAME', '').strip()
+            feature_branch = f"ambient/{session_id}" if session_id else None
+
```
sallyom, line 1345: I'll update  https://github.com/ambient-code/platform/pull/498 to use this branch name - but I think we don't need to always auto-create a branch - we should only auto-create a branch if a user does not provide a branch name in the context modal.

sallyom (commented):
I think we should only auto-generate a new branch if a user does not provide a branch in the context modal. Other than that this LGTM

PR #503, opened 2026-01-10
add pytorch security agents

sumantro93:
Adding Kajuu and Sherru the duo which scans, finds and squashes PyTorch Security Flaws


jeremyeder:
hey, I think we're trying to consolidate agents in https://github.com/opendatahub-io/ai-helpers

PR #521, opened 2026-01-23
fix: message timestamps not being accurate

andybraren:
Still need to test this with an updated backend, but this should add per-message timestamps to the eventstream from AGUI, enabling us to show accurate per-message timestamps. Right now all message timestamps in a session change to the time of the most recently received message.


jeremyeder:
@andybraren check out https://github.com/anthropics/claude-plugins-official/blob/main/plugins/code-simplifier/agents/code-simplifier.md

andybraren:
Timestamps for each message are now passing through the eventstream and appearing in the UI correctly, even after refreshing:

PR #555, opened 2026-01-29
feat: Add setup script for Vertex AI on kind clusters

natifridman:
Adds a helper script to configure the Ambient Code Platform to use Google Cloud Vertex AI instead of the Anthropic API when running locally on kind. 
Uses the same environment variables as Claude Code CLI (GOOGLE_APPLICATION_CREDENTIALS, ANTHROPIC_VERTEX_PROJECT_ID, CLOUD_ML_REGION) for consistency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,14 @@ OPERATOR_IMAGE ?= vteam_operator:latest
 RUNNER_IMAGE ?= vteam_claude_runner:latest
 STATE_SYNC_IMAGE ?= vteam_state_sync:latest
 
+# Vertex AI Configuration (for LOCAL_VERTEX=true)
+# These inherit from environment if set, or can be overridden on command line
+LOCAL_VERTEX ?= false
+ANTHROPIC_VERTEX_PROJECT_ID ?= $(shell echo $$ANTHROPIC_VERTEX_PROJECT_ID)
+CLOUD_ML_REGION ?= $(shell echo $$CLOUD_ML_REGION)
+# Default to ADC location if not set (created by: gcloud auth application-default login)
+GOOGLE_APPLICATION_CREDENTIALS ?= $(or $(shell echo $$GOOGLE_APPLICATION_CREDENTIALS),$(HOME)/.config/gcloud/application_default_credentials.json)
+
 
 # Colors for output (using tput for better compatibility, with fallback to printf-compatible codes)
 # Use shell assignment to evaluate tput at runtime if available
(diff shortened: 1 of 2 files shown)


scripts/setup-vertex-kind.sh
```diff
@@ -0,0 +1,179 @@
...
+echo "Vertex AI is now configured for the kind cluster."
+echo ""
+echo "Configuration applied:"
+echo "  - Namespace: $NAMESPACE"
+echo "  - Project:   $ANTHROPIC_VERTEX_PROJECT_ID"
+echo "  - Region:    $CLOUD_ML_REGION"
+echo ""
+echo "Next steps:"
+echo "  1. Verify the operator started correctly:"
+echo "     kubectl logs -l app=agentic-operator -n $NAMESPACE | grep -i vertex"
```
adalton: Looks like it doesn't generate such a log immediately; you can't use this (at this time) to verify that the operator started correctly.  I think you'd see this only after creating a session.

scripts/setup-vertex-kind.sh
```diff
@@ -0,0 +1,179 @@
...
+echo ""
+
+# Step 2: Patch the operator-config ConfigMap with env vars
+echo "Step 2/3: Patching operator-config ConfigMap..."
+kubectl patch configmap operator-config -n "$NAMESPACE" --type merge -p "{
+  \"data\": {
+    \"CLAUDE_CODE_USE_VERTEX\": \"1\",
+    \"ANTHROPIC_VERTEX_PROJECT_ID\": \"$ANTHROPIC_VERTEX_PROJECT_ID\",
+    \"CLOUD_ML_REGION\": \"$CLOUD_ML_REGION\",
+    \"GOOGLE_APPLICATION_CREDENTIALS\": \"/etc/vertex/ambient-code-key.json\"
```
adalton: Should that be `/app/vertex/...` instead of `/etc/vertex/...`?

scripts/setup-vertex-kind.sh
andybraren: I'm seeing the same - no output when I run this.

scripts/setup-vertex-kind.sh
natifridman: Yes, fixed

scripts/setup-vertex-kind.sh
natifridman: Not sure if you need to wait a bit or if you need to create a new session. I added sleep over there in case it takes some time for the operator

PR #564, opened 2026-01-31
fix: Add missing Switch UI component

jeremyeder:
## Summary
- Add missing Shadcn Switch component required by SDKConfigurationPanel
- Update tsconfig.json with Next.js automatic React JSX runtime configuration

## Background
SDKConfigurationPanel.tsx (committed in 35da252) imports Switch component that was never added to the repository, causing runtime failures when the settings panel is accessed.

## Test Plan
- [x] Frontend builds successfully
- [x] No TypeScript errors
- [x] Switch component follows Shadcn UI patterns

🤖 Generated with Claude Code

### DIFF
--- a/components/frontend/src/components/ui/switch.tsx
+++ b/components/frontend/src/components/ui/switch.tsx
@@ -1,0 +1,35 @@
+"use client"
+
+import * as React from "react"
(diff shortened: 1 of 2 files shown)


jeremyeder:
not sure how this happened but it didnt commit one file of a feature im working on. so here i is.

PR #574, opened 2026-02-03
fix(e2e): Pass ANTHROPIC_API_KEY to Cypress with correct prefix

Gkrumbach07:
## Problem

The E2E test for agent interaction was failing in CI with:
```
Error: ANTHROPIC_API_KEY not set in e2e/.env - agent testing cannot proceed
```

Even though the `ANTHROPIC_API_KEY` GitHub secret was set in the workflow.

## Root Cause

Cypress requires environment variables to use the `CYPRESS_` prefix to be exposed to tests (similar to how `TEST_TOKEN` is passed as `CYPRESS_TEST_TOKEN`).

The `run-tests.sh` script was passing `ANTHROPIC_API_KEY` directly, but Cypress couldn't access it without the proper prefix.

## Changes

1. **`e2e/scripts/run-tests.sh`**: Changed from `ANTHROPIC_API_KEY` to `CYPRESS_ANTHROPIC_API_KEY`
2. **`e2e/cypress.config.ts`**: Updated to read from `CYPRESS_ANTHROPIC_API_KEY` first, then fallback to `ANTHROPIC_API_KEY` for backwards compatibility
3. **`e2e/cypress/e2e/sessions.cy.ts`**: Improved error message to mention both local and CI configuration

## Testing

- [x] Local testing still works (reads from .env)
- [ ] CI testing will work once this PR is merged (reads from GitHub secret)

## Flow

```
CI → ANTHROPIC_API_KEY secret → run-tests.sh → CYPRESS_ANTHROPIC_API_KEY → Cypress → Tests
```

Fixes the failing agent session test in the E2E workflow.

### DIFF
--- a/e2e/scripts/deploy.sh
+++ b/e2e/scripts/deploy.sh
@@ -7,13 +7,6 @@ echo "======================================"
 echo "Deploying Ambient to kind cluster"
 echo "======================================"
 
-# Debug: Check if ANTHROPIC_API_KEY is set (show length, not value)
-if [ -n "${ANTHROPIC_API_KEY:-}" ]; then
-  echo "🔍 Debug: ANTHROPIC_API_KEY is set (length: ${#ANTHROPIC_API_KEY})"
-else
-  echo "🔍 Debug: ANTHROPIC_API_KEY is NOT set in environment"
-fi
-
 # Load .env file if it exists (for ANTHROPIC_API_KEY)
 if [ -f ".env" ]; then
   echo "Loading configuration from .env..."


Gkrumbach07:
## Root Cause Found! 🔍

The issue was more subtle than I initially thought. Here's what was happening:

### The Problem Chain

1. ✅ GitHub secret `ANTHROPIC_API_KEY` **is set** in the repo
2. ✅ Workflow **correctly passes** it to both deploy.sh and run-tests.sh steps
3. ❌ But **deploy.sh creates `.env.test`** with only `TEST_TOKEN` and `CYPRESS_BASE_URL`
4. ❌ Then **run-tests.sh tries to load** ANTHROPIC_API_KEY from `.env` or `.env.local`
5. ❌ Those files **don't exist in CI**, so the secret was never found!

### The Fix

Now `deploy.sh` saves ANTHROPIC_API_KEY to `.env.test` alongside the test token, and `run-tests.sh` loads it from there.

### Flow Now

```
GitHub Secret → deploy.sh → .env.test → run-tests.sh → CYPRESS_ANTHROPIC_API_KEY → Cypress → Tests ✅
```

The next CI run should pass the agent session test! 🎉

Gkrumbach07:
## 🎯 Real Root Cause Found!

The issue is that **this PR is from a fork** (`isCrossRepository: true`).

### Why The Test Was Failing

**GitHub Actions does NOT expose secrets to pull requests from forks** - this is a deliberate security feature to prevent malicious PRs from exfiltrating secrets.

So even though:
- ✅ The `ANTHROPIC_API_KEY` secret exists in `ambient-code/platform`
- ✅ The workflow declares `env: ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}`
- ✅ Our scripts correctly pass it to Cypress

GitHub Actions **intentionally doesn't pass the secret value** because the PR is from `Gkrumbach07/platform` → `ambient-code/platform`.

### The Solution

The test now **gracefully skips** when ANTHROPIC_API_KEY is unavailable instead of throwing an error:

```typescript
if (!apiKey) {
  cy.log('⚠️ ANTHROPIC_API_KEY not available - skipping agent interaction test')
  cy.log('   This is expected for PRs from forks (GitHub security restriction)')
  this.skip()  // Skip instead of throw
}
```

### What This Means

- ✅ **Fork PRs**: E2E tests pass (agent test skipped) 
- ✅ **Main repo PRs**: E2E tests pass with full agent interaction
- ✅ **Local dev**: E2E tests pass with full agent interaction (if `.env` has key)

The CI should now pass! 🎉

Gkrumbach07:
## ✨ Simplified to Single Workflow!

Changed approach to something much simpler:

### How It Works Now

**ONE workflow** that ONLY runs when the `safe-to-test` label is added to a PR.

### Security Model

1. ✅ Maintainer reviews PR code (manual security check)
2. ✅ Maintainer adds `safe-to-test` label (explicit approval)
3. ✅ Workflow runs with full secrets (`pull_request_target` context)
4. ✅ Tests PR code (builds images from PR branch)
5. ✅ Posts results as PR comment

### Usage

```bash
# Maintainer adds label to approved PR
gh pr edit <PR_NUMBER> --add-label safe-to-test

# Workflow runs automatically
# Results posted to PR
```

### Benefits

- 🎯 **Simple**: One workflow, one trigger
- 🔒 **Secure**: Maintainer approval required
- 💯 **Complete**: Full test suite with secrets
- 📝 **Clear**: Docs in `docs/testing/e2e-testing.md`

No more automatic runs, no fork confusion, no split workflows. Just explicit maintainer approval! 🎉

To test this PR: Add the `safe-to-test` label and watch it run!

Gkrumbach07:
🚀 **Label added - E2E tests starting!**

The `safe-to-test` label has been added, triggering the E2E workflow with full secrets.

Watch the workflow run here: https://github.com/ambient-code/platform/actions/workflows/e2e.yml

Expected results:
- ✅ All 12 tests should pass (including agent interaction)
- ✅ Results will be posted as a comment when complete

This verifies the complete fix works end-to-end! 🎯

Gkrumbach07:
## 🔧 Fixed Permissions Issue

The workflow was failing with "Resource not accessible by integration" because `pull_request_target` needs explicit permissions to comment on PRs.

### What I Fixed

Added permissions block to workflow:
```yaml
permissions:
  contents: read
  pull-requests: write
  issues: write
```

Also made comment steps non-fatal (`continue-on-error: true`) so they don't break the actual tests if commenting fails.

### Re-triggered

Removed and re-added the `safe-to-test` label to trigger a new run with the fix.

Watch here: https://github.com/ambient-code/platform/actions/workflows/e2e.yml

Should work properly this time! 🤞

Gkrumbach07:
✅ Cleaned up workflow - removed PR comment steps.

Results now show up cleanly in GitHub Actions UI and PR checks section.

Re-triggered: https://github.com/ambient-code/platform/actions/workflows/e2e.yml

PR #575, opened 2026-02-03
fix(e2e): simplify E2E workflow to trigger only on safe-to-test label

Gkrumbach07:
## Problem

E2E tests were failing in CI because fork PRs don't have access to GitHub secrets.

## Solution

Simplified to ONE workflow that ONLY triggers when `safe-to-test` label is added.

Uses `pull_request_target` for secret access with explicit maintainer approval.

## How to Use

```bash
gh pr edit <PR_NUMBER> --add-label safe-to-test
```

## Changes

- Trigger: `pull_request_target` with label check
- Security: Explicit approval via label
- Tests: Full suite with secrets
- Clean: No PR comments

## Benefits

🔒 Secure - maintainer approval required
🎯 Simple - one workflow, one trigger  
💯 Complete - full agent interaction testing

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -9,12 +9,6 @@ on:
     types: [labeled]
     branches: [ main, master ]
 
-# Required for pull_request_target to comment on PRs
-permissions:
-  contents: read
-  pull-requests: write
-  issues: write
-
 concurrency:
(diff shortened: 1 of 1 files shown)


Gkrumbach07:
✅ Rebased on `upstream/main`

Branch is now up to date with main. The `safe-to-test` label is already added, so E2E tests should be running!

PR #604, opened 2026-02-10
feat(operator): externalize session container resource limits to Conf…

jeremyeder:
**Jira:** [RHOAIENG-51900](https://issues.redhat.com/browse/RHOAIENG-51900)

This has to get rolled to both clusters. It's basic hygiene.
I pulled a must-gather and did an analysis. This change fixes almost everything.

@Gkrumbach07: 

Move all session container resource requests/limits to env vars loaded from the operator-config ConfigMap, following the existing pattern used by AMBIENT_CODE_RUNNER_IMAGE and STATE_SYNC_IMAGE. Add resources to the ambient-content container (previously had none). Add a LimitRange guardrail for session namespaces to enforce floor values on any new container.

wdyt

### DIFF
--- a/components/manifests/base/kustomization.yaml
+++ b/components/manifests/base/kustomization.yaml
@@ -15,6 +15,7 @@ resources:
 - public-api-deployment.yaml
 - workspace-pvc.yaml
 - minio-deployment.yaml
+- session-limitrange.yaml
 
 # Default images (can be overridden by overlays)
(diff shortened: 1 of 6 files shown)


jeremyeder:
We might still need this some day but we've gotten along without it for now. It's one of the things that we would need to tackle to make system changes transparent to users.

PR #606, opened 2026-02-10
feat: add session-config repo support (spec.configRepo)

jeremyeder:


components/runners/state-sync/hydrate.sh
```diff
@@ -204,6 +204,51 @@ else
...
+        # cp -rn (no-clobber, GNU coreutils) ensures workspace repo files take precedence
+        cp -rn "$CONFIG_TEMP"/. /workspace/ 2>/dev/null || true
+
+        rm -rf "$CONFIG_TEMP"
+        echo "  ✓ Session config applied to workspace"
+    else
+        echo "  ✗ Failed to clone session-config repository after 2 attempts"
+        echo "Failed to clone session-config repository: $CONFIG_REPO_GIT_URL (branch: $CONFIG_BRANCH)" > /workspace/.config-repo-error
+        # Non-fatal: session continues without config overlay
+    fi
```
kahowell, line 247: Instead of doing this, can we map this to claude "user" settings, and land them in `$HOME/.claude`? We can then document that session config is implemented as user config and follows claude standards for user config.

jeremyeder:
Needs a greenfield design.

PR #612, opened 2026-02-10
feat: dynamic model support via MODELS_JSON ConfigMap

jeremyeder:
**Jira:** [RHOAIENG-51899](https://issues.redhat.com/browse/RHOAIENG-51899)

## Summary

- Add `MODELS_JSON` key to the existing `operator-config` ConfigMap (all overlays: kind, e2e, production, local-dev, minikube)
- Backend parses `MODELS_JSON` env var and includes `models` array in `GET /api/cluster-info` response
- Frontend reads models from `useClusterInfo()` hook and populates the session creation dropdown dynamically (falls back to hardcoded list if empty)
- Operator passes `MODELS_JSON` through to runner pods as an env var
- Runner builds `VERTEX_MODEL_MAP` from `MODELS_JSON` at startup (falls back to hardcoded map)

Adding or removing a model is now: edit ConfigMap → `kubectl rollout restart` → done. No code changes or image rebuilds needed.

## Test plan

- [ ] `GET /api/cluster-info` returns `models` array populated from ConfigMap
- [ ] Frontend model dropdown shows models from API response
- [ ] Default model selection respects `default: true` flag
- [ ] Frontend falls back to hardcoded list when `models` is empty
- [ ] Runner resolves Vertex model IDs from `MODELS_JSON` env var
- [ ] Runner falls back to hardcoded map when `MODELS_JSON` is unset
- [ ] Edit ConfigMap + rollout restart updates models without image rebuilds
- [ ] `go vet`, `gofmt`, `tsc --noEmit`, `next build` all pass

### DIFF
--- a/components/backend/handlers/projects.go
+++ b/components/backend/handlers/projects.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"log"
 	"net/http"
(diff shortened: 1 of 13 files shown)


jeremyeder:
michael is doing this. duplicate.

PR #628, opened 2026-02-14
fix: use full UBI9 nodejs image for frontend build stage

jeremyeder:
## Summary

- Switch the frontend Dockerfile build stage from `ubi9/nodejs-20-minimal` to the full `ubi9/nodejs-20` image

## Problem

Next.js 16 Turbopack requires native SWC binaries that depend on glibc. The `ubi9/nodejs-20-minimal` image lacks glibc, which causes arm64 build failures when Turbopack attempts to load its native bindings during the build stage.

## Fix

Use the full `ubi9/nodejs-20` image for the frontend build stage, which includes glibc and all required system libraries for SWC/Turbopack native binaries to load correctly on both amd64 and arm64 architectures.

### DIFF
--- a/components/frontend/Dockerfile
+++ b/components/frontend/Dockerfile
@@ -10,7 +10,9 @@ COPY package.json package-lock.json* ./
 RUN npm ci
 
 # Rebuild the source code only when needed
-FROM registry.access.redhat.com/ubi9/nodejs-20-minimal AS builder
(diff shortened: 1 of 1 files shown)


jeremyeder:
I hit this for the first time today, on my mac. This change seems fine.

jeremyeder:
Minimal=76mb / 211mb uncompressed
Full=217mb / 585mb uncompressed

PR #631, opened 2026-02-15
feat: Add resizable and collapsible left panel to workspace page

jeremyeder:
It looks mostly similar to the Sessions page version. Good enough anyway.

Implements left panel improvements for the workspace/sessions list page:
- Users can hide the sidebar by clicking "Hide sidebar" button at the bottom
- Users can show the sidebar by clicking floating chevron button on left edge
- Users can resize the sidebar width by dragging the resize handle
- Panel visibility and width preferences are persisted to localStorage

Follows the same pattern as the session detail page implementation.

Fixes: RHOAIENG-45393

### DIFF
--- a/components/frontend/src/app/projects/[name]/page.tsx
+++ b/components/frontend/src/app/projects/[name]/page.tsx
@@ -2,7 +2,7 @@
 
 import { useState, useEffect } from 'react';
 import { useParams, useSearchParams } from 'next/navigation';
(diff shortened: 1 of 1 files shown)


markturansky:
LGTM.

how might be validate/test something like this?  a screenshot from you showing the UI working? This change is pretty trivial.

Claude through Chromium can control browsers now, too.

PR #645, opened 2026-02-16
RHOAIENG-51548: fix: replace dev-* Makefile targets with local-*

mprpic:
The CRC-based dev-* targets (dev-logs, dev-start, dev-stop, etc.) were replaced with local-* equivalents in 5553056 but docs were never updated to reflect this changes.


mprpic:
Closing in favor of https://github.com/ambient-code/platform/pull/784

PR #656, opened 2026-02-18
perf(operator): parallelize secret reads, remove monitorPod goroutine…

syntaxsdev:
- Replace sequential secret/config reads with errgroup-based parallel fetches, reducing API round-trips when creating sessions under load
- Remove monitorPod goroutine-per-session pattern; controller-runtime reconciler handles pod status via its watch/requeue loop
- Eliminate redundant CR re-GET in ReconcilePendingSession; use the object already provided by the controller-runtime informer cache
- Switch gauge callbacks in otel_metrics.go to use the informer cache instead of a cluster-wide LIST against the API server every 30s
- Replace clearAnnotation GET+UPDATE (2 API calls) with a single JSON merge patch


syntaxsdev:
hold off

PR #651, opened 2026-02-18
feat: auto-stop idle sessions and preserve git repo state across restarts

adalton:
## Summary

  - Add configurable inactivity timeout that automatically stops idle sessions, reclaiming cluster resources
  - Preserve git repo state (local branches, uncommitted/staged changes) to S3 on pod shutdown and restore on resume, so work is never lost when pods are stopped or recycled
  - Three-tier timeout resolution: session-level `spec.inactivityTimeout` > project-level `ProjectSettings.spec.inactivityTimeoutSeconds` > default 24h; set to 0 to disable

  ## What changed

  **CRDs:**
  - `AgenticSession`: added `spec.inactivityTimeout`, `status.lastActivityTime`, `status.stoppedReason`
  - `ProjectSettings`: added `spec.inactivityTimeoutSeconds`

  **Backend:**
  - Debounced activity tracking in `agui_proxy.go` updates `status.lastActivityTime` on AG-UI events (RUN_STARTED, TEXT_MESSAGE_START, TEXT_MESSAGE_CONTENT, TOOL_CALL_START)
  - `parseStatus()` extracts new status fields for API responses

  **Operator:**
  - New `inactivity.go` with `shouldAutoStop()`, `resolveInactivityTimeout()`, `triggerInactivityStop()`, and per-namespace ProjectSettings cache
  - `monitorPod()` checks inactivity on each tick; re-reads CR before stopping to avoid races
  - `reconciler.go` reads `stop-reason` annotation to set `status.stoppedReason` and condition reason

  **Frontend:**
  - "Stopped (idle)" badge on session cards
  - Inactivity alert banner in session header (full/actions-only modes)
  - "This session was automatically stopped after being idle" message in hibernated section

  **State-sync:**
  - `sync.sh`: on SIGTERM, creates git bundles, uncommitted/staged patches, and metadata.json per repo
  - `hydrate.sh`: restores repos from bundles, checks out saved branch, applies patches (best-effort)
  - `TerminationGracePeriodSeconds` increased from 30 to 60

  **Tests:**
  - `inactivity_test.go`: 16 tests covering `shouldAutoStop`, `resolveInactivityTimeout`, `triggerInactivityStop`, `getProjectInactivityTimeout`
  - `agui_proxy_test.go`: 18 subtests for `isActivityEvent`

  ## Test plan

  - [x] Operator and backend unit tests pass (`go test -race ./...`)
  - [x] Frontend builds with zero errors/warnings (`npm run build`)
[...]

PR #658, opened 2026-02-19
fix: replace DAC_READ_SEARCH with RunAsUser for OpenShift SCC compatibility

Gkrumbach07:
## Summary

- Replace `DAC_READ_SEARCH` capability on the state-sync sidecar with `RunAsUser: 1001` to match the runner container UID
- Fixes pod creation failures on OpenShift where `restricted-v2` SCC blocks `DAC_READ_SEARCH`

## Root Cause

PR #651 introduced `DAC_READ_SEARCH` on the state-sync container so it could read workspace files owned by the runner (UID 1001, mode 700). This works on kind/Kubernetes but fails on OpenShift because `restricted-v2` does not allow adding that capability.

## Fix

Running state-sync as the same UID (1001) as the runner gives it native read access to the workspace files without any elevated capabilities. Fully compatible with `restricted-v2` and all OpenShift SCCs.

## Test plan

- [x] `go build ./...` passes for operator
- [ ] Deploy on OpenShift and verify sessions create pods successfully
- [ ] Verify state-sync can still read workspace files and sync to S3

### DIFF
--- a/components/operator/internal/handlers/sessions.go
+++ b/components/operator/internal/handlers/sessions.go
@@ -1146,10 +1146,6 @@ func handleAgenticSessionEvent(obj *unstructured.Unstructured) error {
 				SecurityContext: &corev1.SecurityContext{
 					AllowPrivilegeEscalation: boolPtr(false),
 					ReadOnlyRootFilesystem:   boolPtr(false),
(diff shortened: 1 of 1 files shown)


ktdreyer:
Wondering if there is a straightforward way to apply this to `stage` ahead of merging

PR #683, opened 2026-02-25
fix: run ambient-api-server-db as postgres user (UID 999)

jeremyeder:
## Summary

- Add `runAsUser: 999`, `runAsGroup: 999`, `fsGroup: 999` to the ambient-api-server-db deployment
- The `postgres:16.2` Docker image entrypoint tries to `chown`/`chmod` directories when running as root, which fails under the restricted securityContext (`capabilities.drop: ALL`)
- Running as UID 999 (postgres) makes the entrypoint skip those privileged operations, and `fsGroup` ensures the PVC mount is group-writable

## Test plan

- [x] Verified fix resolves CrashLoopBackOff on kind cluster (`make kind-up`)
- [ ] Verify on minikube (`make local-up`)
- [ ] Verify on OpenShift (production overlay)

### DIFF
--- a/components/manifests/base/ambient-api-server-db.yml
+++ b/components/manifests/base/ambient-api-server-db.yml
@@ -54,6 +54,10 @@ spec:
         app: ambient-api-server
         component: database
     spec:
+      securityContext:
+        runAsUser: 999
+        runAsGroup: 999
+        fsGroup: 999
       containers:
         - name: postgresql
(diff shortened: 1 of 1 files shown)


markturansky:
@jeremyeder this LGTM but there are no automated tests yet for the DB, AFAICT.  something like https://github.com/ambient-code/platform/pull/699 adds integration tests that use the db.

PR #692, opened 2026-02-25
feat: add automated model discovery with feature-flag-gated availability

maskarb:
## Summary

Single-source model manifest (`models.json`) with feature-flag-gated availability. Models are defined once in the manifest, discovered automatically via GitHub Action, and gated per-workspace through Unleash feature flags with admin override support.

### Model endpoint (project-scoped)

`GET /api/projects/:projectName/models` returns the filtered model list for a workspace:

1. Loads the model manifest from the `ambient-models` ConfigMap (mounted volume)
2. For each model (skipping unavailable ones):
   - **Default model**: always included
   - **Others**: checks the workspace's `feature-flag-overrides` ConfigMap first (via user-scoped K8s client); if no override exists, falls back to Unleash SDK evaluation of `model.<id>.enabled`
3. Returns only models that pass filtering
4. On manifest read failure, falls back to a cached copy (atomic.Pointer) to preserve flag semantics; only uses hardcoded defaults on cold start with no cache

The global `/api/models` endpoint has been removed.

### Flag sync at startup

The backend runs `SyncModelFlagsAsync` at startup to ensure Unleash flags exist for all manifest models:

1. Reads the mounted `models.json` manifest
2. For each available, non-default model:
   - `GET /api/admin/projects/{project}/features/{flagName}` — check if flag exists
   - If not: `POST /api/admin/projects/{project}/features` — create flag (disabled, type `release`)
   - `POST /api/admin/features/{flagName}/tags` — add `scope: workspace` tag
   - `POST /api/admin/projects/{project}/features/{flagName}/environments/{env}/strategies` — add `flexibleRollout` strategy at 0%
3. Logs summary: created, skipped, excluded (default + unavailable), errors
4. Retries up to 3 times with 10s delay; accepts `context.Context` for cancellation on shutdown
5. All URL path segments are escaped via `url.PathEscape()`

### Feature Flags admin UI overrides

Replaced the toggle-based UI with a tri-state override control per flag:

| Column | Description |
|--------|-------------|
| **Default** | On/Off badge showing the Unleash platform-level state for the flag |
| **Override** | Segmented control: **Default** (inherit platform) · **On** (force enable) · **Off** (force disable) |

Overrides are stored in the workspace's `feature-flag-overrides` ConfigMap and take precedence over Unleash evaluation in the models endpoint. Saving overrides invalidates both the feature-flags and models query caches so changes are reflected immediately.

### Session creation model validation

`CreateSession` validates the requested model via `isModelAvailable`, which checks:
1. Empty model ID → allowed (uses default)
2. Manifest lookup with cached fallback → fail-open on cold start only
3. Default model → always available
4. Available flag + workspace overrides (via user-scoped K8s client) + Unleash fallback

### Other changes

- **GitHub Action**: removed disabled flag-creation step; updated PR body template
- **Python script**: deleted `scripts/create-model-flags.py` (replaced by Go subcommand)
- **Frontend proxy**: added Next.js API route handler at `src/app/api/projects/[name]/models/route.ts`
- **Frontend**: model dropdown fetches from API on dialog open; query invalidated after flag override saves
- **Manifest**: `models.json` updated (removed test model)
- **Reverted**: unrelated `ambient-api-server` go.mod/go.sum dependency bumps

## Test plan

- [ ] Deploy to cluster and verify init sync logs show correct counts (created + skipped + excluded = total models)
- [ ] Verify model dropdown only shows default model + Unleash-enabled models
- [ ] Set a workspace override to "On" for a disabled model → verify it appears in the dropdown
- [ ] Set a workspace override to "Off" for an enabled model → verify it disappears from the dropdown
- [ ] Reset override to "Default" → verify it falls back to Unleash evaluation
- [ ] Verify Feature Flags admin page shows default state badges and override controls
- [ ] Verify model dropdown updates immediately after saving flag overrides (no 60s stale window)
- [ ] Verify CreateSession rejects a model that is disabled by workspace override

### DIFF
--- a/scripts/model-discovery.py
+++ b/scripts/model-discovery.py
@@ -60,7 +60,10 @@ def get_access_token() -> str:
             capture_output=True,
             text=True,
             check=True,
+            timeout=30,
         )
+    except subprocess.TimeoutExpired:
+        raise RuntimeError("Timed out getting GCP access token via gcloud")
     except subprocess.CalledProcessError:
         raise RuntimeError("Failed to get GCP access token via gcloud")
     return result.stdout.strip()


components/manifests/base/models.json
jeremyeder, line 1: Should we bother pruning this list? If so, perhaps keep the latest 2 versions of each model family. e.g. opus 4.5 and 4.6, but NOT 4.1.

jeremyeder: dont bother pruning to start....

scripts/model-discovery.py
```diff
@@ -0,0 +1,286 @@
...
+)
+
+# Known Anthropic model base names. Add new models here as they are released.
+# Version resolution and availability probing are automatic.
+KNOWN_MODELS = [
+    "claude-sonnet-4-6",
+    "claude-sonnet-4-5",
+    "claude-opus-4-6",
+    "claude-opus-4-5",
+    "claude-opus-4-1",
```
jeremyeder: 👀

scripts/model-discovery.py
```diff
@@ -0,0 +1,286 @@
...
+    "claude-haiku-4-5",
+]
+
+# Tier classification: models matching these patterns are "standard", others "premium"
+STANDARD_TIER_PATTERNS = ["sonnet", "haiku"]
```
jeremyeder: what is this? why?

scripts/model-discovery.py
```diff
@@ -0,0 +1,286 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def get_access_token() -> str:
```
jeremyeder, line 55: What secret does this use?

scripts/model-discovery.py
```diff
@@ -0,0 +1,286 @@
...
+    return "premium"
+
+
+def probe_model(region: str, project_id: str, vertex_id: str, token: str) -> str:
+    """Probe a Vertex AI model endpoint.
+
+    Returns:
+        "available"   - 200 or 400 (model exists, endpoint responds)
+        "unavailable" - 404 (model not found)
+        "unknown"     - any other status (transient error, leave unchanged)
```
jeremyeder, line 140: some hardening/retry here please.

scripts/model-discovery.py
maskarb: This will be utilized by the github action:
https://github.com/ambient-code/platform/pull/692/changes#diff-35c50ad1ffcfc8f3012e7bf931f3e9b85002ebe25ef62e304529ce626a3dead6R30-R43

we'll need to set up variables in the repo settings if they dont already exist.

scripts/model-discovery.py
maskarb: This isn't strictly required for anything. It's just another classification we can fallback to if we ever wanted to limit users to "standard" or "premium" models. We can remove this for now and implement in the future if we ever wanted that.

maskarb:
closed in favor of #772

PR #693, opened 2026-02-25
fix: timestamps changing

jwm4:
Fixes the issue with the timestamps changing so they match each other.  I have verified manually in Kind.


jwm4:
FWIW, I agree with Claude Code that we should have a unit testing framework for the front end, but I think that should be its own work item.

PR #701, opened 2026-02-26
improvement: Extract amber.review command

markturansky:
improvement: creates reusable amber review command that's called from git workflow


markturansky:
use https://github.com/ambient-code/platform/pull/707

PR #728, opened 2026-02-27
RHOAIENG-51372: add smoke test workflow for preprod runner

ktdreyer:
## Summary
- Adds a manual `workflow_dispatch` workflow to verify the standalone GitHub Actions runner on preprod can pick up jobs. See [RHOAIENG-51372](https://issues.redhat.com/browse/RHOAIENG-51372) for context.
- Just prints the hostname — meant to be triggered once and then removed

## Test plan
- Merge, then trigger from Actions tab → ["TEST: preprod runner"](https://github.com/ambient-code/platform/actions/workflows/test-preprod-runner.yml) → Run workflow
- Confirm the job runs on the preprod self-hosted runner

### DIFF
--- a/.github/workflows/test-preprod-runner.yml
+++ b/.github/workflows/test-preprod-runner.yml
@@ -1,0 +1,12 @@
+name: "TEST: preprod runner"
+on:
+  push:
+    paths:
+      - .github/workflows/test-preprod-runner.yml
+  workflow_dispatch:
+
+jobs:
+  hello:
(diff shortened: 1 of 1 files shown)


ktdreyer:
We don't need this in `main` - only for `workflow_dispatch`. I found that I can just trigger on `push:`. ([example](https://github.com/ambient-code/platform/actions/runs/22504936344/job/65201140905))

PR #759, opened 2026-03-02
docs: Ambient ERD and RBAC Software Factory proposal

markturansky:
**Jira:** [RHOAIENG-51884](https://issues.redhat.com/browse/RHOAIENG-51884)

## Summary

- Adds `docs/internal/design/ambient-erd.md` — canonical Mermaid ERD of the current Ambient data model, derived directly from `openapi.yaml`. This is the **living spec**: future Kind additions are proposed as diffs against it.
- Adds `docs/internal/proposals/rbac-rolebinding.md` — proposal to add first-class RBAC using the Software Factory pattern.

## Design Decisions

**Static vs Dynamic:**
- `Permission` and `Role` are static Go constants/structs — compile-time safety, TDD-friendly, no DB table
- `RoleBinding` is the only new database-backed Kind (one TRex generator invocation)

**Three binding scopes:**
- Global (`project_id = null, session_id = null`)
- Workspace (`project_id = <id>, session_id = null`)
- Session (`project_id = <id>, session_id = <id>`)

**API-native enforcement** — permission middleware in `ambient-api-server` only; no Kubernetes RBAC sync required.

## How to use these docs

1. Review `ambient-erd.md` as the current state spec
2. Review `rbac-rolebinding.md` as the proposed diff (adds `RoleBinding` Kind, removes `group_access` JSON blob)
3. If accepted, update `ambient-erd.md` to include `RoleBinding` and the Software Factory executes the cascade

### DIFF
--- a/docs/internal/design/ambient-erd.md
+++ b/docs/internal/design/ambient-erd.md
@@ -1,0 +1,122 @@
+# Ambient Platform — Data Model ERD
+
+This is the canonical entity-relationship diagram for the `ambient-api-server`. It reflects the current OpenAPI spec (`platform-api-server/components/ambient-api-server/openapi/openapi.yaml`) and the backing PostgreSQL models.
+
+All Kinds inherit `id` (KSUID), `kind`, `href`, `created_at`, `updated_at`, and `deleted_at` from `api.Meta` (TRex base). These are omitted from individual entity fields below for clarity but are present on every record.
+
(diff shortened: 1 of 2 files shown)


jeremyeder:
This feature falls into what we might consider a "core" feature of ACP. I propose we start picking up spec-kit opportunistically for "core" work. This could be the first one to do it. We have to agree together. I added a note for standup.

markturansky:
A spec like this was designed and used for #866.  There might not be an official "spec" for us as a feature. It seems like it's just good documentation that becomes a spec.

Issue #749, opened 2026-03-01
Security: amber-auto-review.yml workflow is vulnerable to prompt injection from fork PRs


varunsh-coder:
## Summary

Your `amber-auto-review.yml` workflow is configured in a way that allows any GitHub user to trigger an AI code review with write permissions on untrusted fork code. This was recently exploited by the `hackerbot-claw` bot (PRs #732 and #733), which attempted prompt injection via a poisoned `CLAUDE.md`. Claude detected and refused the attack — but the underlying workflow configuration that enabled the attack vector is still in place.

A fix was applied on March 1 (commit `ed18288`) but was [reverted 24 minutes later](https://github.com/ambient-code/platform/pull/743) in the belief that the replacement was accidental. The revert restored the vulnerable configuration.

## Vulnerable Configuration

The workflow at [`.github/workflows/amber-auto-review.yml`](https://github.com/ambient-code/platform/blob/main/.github/workflows/amber-auto-review.yml) has four issues that combine to create the attack surface:

1. **`pull_request_target` trigger** — Runs with the base repository's secrets and permissions, even for fork PRs. The [official `claude-code-action` documentation](https://github.com/anthropics/claude-code-action) uses `pull_request` (not `pull_request_target`) in every example.

2. **Fork code checkout** — The workflow checks out `github.event.pull_request.head.repo.full_name` at `github.event.pull_request.head.ref`, which means it loads the attacker's code — including any modified `CLAUDE.md` — as trusted context.

3. **`allowed_non_write_users: '*'`** — Allows any GitHub user to trigger the workflow. The [claude-code-action security documentation](https://github.com/anthropics/claude-code-action/blob/main/docs/security.md) explicitly warns this is "a significant security risk."

4. **`contents: write` permissions** — Grants the workflow token write access to repository contents. The official docs recommend `contents: read`.

## What Happened

On February 28, 2026, `hackerbot-claw` opened PRs #732 and #733, replacing `CLAUDE.md` with social engineering instructions designed to manipulate Claude into committing unauthorized changes. The workflow ran, checked out the fork's code, and loaded the poisoned `CLAUDE.md` as trusted project context.

Claude (running `claude-sonnet-4-6`) detected both prompt injection attempts and refused to comply. The tool allowlisting (`--allowedTools` restricting Claude to read-only `gh` commands) provided an additional layer of defense.

## Recommended Fixes

[...]

PR #763, opened 2026-03-02
RHOAIENG-51481: switch pull_request_target to pull_request

ktdreyer:
External contributors disclosed vulnerabilities in our `pull_request_target` workflows. Switching to `pull_request` prevents untrusted fork code from running with access to repository secrets.

Leave the dependabot-auto-merge workflow unchanged, because it checks `github.actor` as a guard and never checks out code.

Once we merge this, these workflows will no longer act on forks.

```
Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
```

Fixes: #749 

https://issues.redhat.com/browse/RHOAIENG-51481

### DIFF
--- a/.github/workflows/components-build-deploy.yml
+++ b/.github/workflows/components-build-deploy.yml
@@ -12,7 +12,7 @@ on:
       - 'components/frontend/**'
       - 'components/public-api/**'
       - 'components/ambient-api-server/**'
-  pull_request_target:
+  pull_request:
     branches: [main]
     paths:
       - '.github/workflows/components-build-deploy.yml'
@@ -42,6 +42,7 @@ concurrency:
 
 jobs:
   detect-changes:
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     outputs:
       frontend: ${{ steps.filter.outputs.frontend }}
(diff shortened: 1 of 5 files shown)


.github/workflows/components-build-deploy.yml
```diff
@@ -166,13 +166,13 @@ jobs:
...
 
       - name: Build ${{ matrix.component.name }} image for pull requests but don't push
-        if: (matrix.component.changed == 'true' || github.event.inputs.force_build_all == 'true' || contains(github.event.inputs.components, matrix.component.name)) && github.event_name == 'pull_request_target'
+        if: (matrix.component.changed == 'true' || github.event.inputs.force_build_all == 'true' || contains(github.event.inputs.components, matrix.component.name)) && github.event_name == 'pull_request'
         uses: docker/build-push-action@v6
         with:
           context: ${{ matrix.component.context }}
           file: ${{ matrix.component.dockerfile }}
           platforms: linux/amd64,linux/arm64
-          push: true
```
ktdreyer, line 175: per gage, this is wrong. Claude was operating on a stale comment.

ktdreyer:
Long-term, we want a "safe-to-test" label that operates on the exact sha that is safe to test, rather than blindly trust-on-first-use

.github/workflows/components-build-deploy.yml
ktdreyer: #764 tracks fixing that stale comment separately (lower-priority)

ktdreyer:
Upon reflection, these workflows will likely run-and-fail now, because we're not adding guards "don't run these PRs from forks at all". I will add conditionals so we only execute these `pull_requests` if the author is a member of the org (in other words: they have permission to push here).

EDIT: I've done this now.

PR #786, opened 2026-03-03
fix: Minimize old Claude review comments errors

maskarb:
((COUNT++)) returns exit code 1 when COUNT is 0 because post-increment evaluates to the previous value (0 = falsy). With set -e, this kills the step after successfully minimizing the first comment. Use COUNT=$((COUNT + 1)) which always returns exit code 0.


markturansky (approved):
LGTM

PR #805, opened 2026-03-04
feat(RHOAIENG-50753): add LDAP user and group autocomplete for workspace sharing

mprpic:
Integrate LDAP to provide autocomplete when sharing workspaces. The backend queries LDAP with substring matching on uid, givenName, and sn for users, and prefix matching on cn for groups. Results are cached in-memory with a 5-minute TTL. The frontend replaces the plain text input with an autocomplete dropdown that debounces queries and supports keyboard navigation. LDAP is optional and gracefully degrades to manual input when unconfigured or unreachable.

**Jira:** [RHOAIENG-50753](https://issues.redhat.com/browse/RHOAIENG-50753)

### DIFF
--- a/components/frontend/src/components/workspace-sections/__tests__/sharing-section.test.tsx
+++ b/components/frontend/src/components/workspace-sections/__tests__/sharing-section.test.tsx
@@ -26,6 +26,10 @@ vi.mock('@/services/queries', () => ({
   })),
 }));
 
+vi.mock('@/services/queries/use-feature-flags-admin', () => ({
+  useWorkspaceFlag: vi.fn(() => ({ enabled: false })),
+}));
+
 vi.mock('@/hooks/use-toast', () => ({
   successToast: vi.fn(),
   errorToast: vi.fn(),


mprpic:
Disabling TLS verification in production is intentional for now until we switch to an authenticated LDAP server. The risk of doing this is minimal since the cluster is for authenticated users only and the risk of MITM is low.

components/manifests/overlays/production/ldap-config.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ldap-config
+  labels:
+    app: backend-api
+data:
+  LDAP_URL: "ldaps://ldap.corp.redhat.com"
+  LDAP_BASE_DN: "ou=users,dc=redhat,dc=com"
+  LDAP_SKIP_TLS_VERIFY: "true"
```
ktdreyer, line 10: Do we need `LDAP_SKIP_TLS_VERIFY`?

mprpic: Yes, we don't have CA certs installed yet since we have to be on the internal network to pull them in to the images at build time.

ktdreyer: We don't want to bake these certs into any image. But there's gotta be a best practice for inserting the RH CA certs into the image at runtime.

For years I've been doing this in apps and CI jobs:

```
sudo curl https://certs.corp.redhat.com/certs/Current-IT-Root-CAs.pem -o /etc/pki/ca-trust/source/anchors/Current-IT-Root-CAs.pem
sudo update-ca-trust extract
```

I'm an openshift noob, and we should be able to use a `ConfigMap` here. Maybe IT has some best practice for this too?

mprpic: Yea, the best practice is to just add those two instructions into the Containerfile so that the cert is fetched at build time. This too assumes though that those builds happen on the internal network where you have access to pull the correct certs. I think that is the case once the cluster is migrated to MP+, right?

ktdreyer: I would not consider that a best practice. It's common to bake the internal certs into images, but it's not a good practice to do that or distribute those images.

There must be an openshift-friendly way to do this.

mprpic: My assumption is that the images are built and deployed to an internal registry only, but checking now, we do push everything to https://quay.io/organization/ambient_code.

So yea, adding the certs to a ConfigMap and then mounting that to `/etc/pki/ca-trust/extracted/pem` is probably best.

I'd like to add that in a separate PR that will also switch to ipa.redhat.com instead of the corp instance.

Gkrumbach07:
## Review Queue Status

| Check | Status | Detail |
|-------|--------|--------|
| CI | pass | --- |
| Conflicts | FAIL | Has merge conflicts with main |
| Reviews | FAIL | CHANGES_REQUESTED from @ktdreyer — LDAP cert injection approach needs resolution |
| Jira | pass | RHOAIENG-50753 |

**Action needed:** Rebase to resolve conflicts. Discuss LDAP cert injection approach with @ktdreyer (runtime cert mounting vs baking into image).

> Auto-generated by Review Queue workflow. Updated when PR changes.

mprpic:
@mergifyio queue

PR #812, opened 2026-03-04
fix: allow bot actors in amber auto-review workflow

Gkrumbach07:
## Summary

- Adds `allowed_bots: '*'` to the `anthropics/claude-code-action@v1` step so the amber code review runs on PRs opened by bots (dependabot, renovate, etc.)
- Previously the workflow would fail with: `Action failed with error: Workflow initiated by non-human actor: dependabot (type: Bot)`

## Test plan

- [ ] Open or sync a dependabot PR and verify the amber-review job completes successfully instead of failing

### DIFF
--- a/.github/workflows/amber-auto-review.yml
+++ b/.github/workflows/amber-auto-review.yml
@@ -93,7 +93,7 @@ jobs:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           github_token: ${{ secrets.GITHUB_TOKEN }}
           allowed_non_write_users: '*'
-          allowed_bots: '*'
+          allowed_bots: 'dependabot[bot],github-actions[bot]'
           claude_args: |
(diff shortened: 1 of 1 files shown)


.github/workflows/amber-auto-review.yml
```diff
@@ -93,6 +93,7 @@ jobs:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           github_token: ${{ secrets.GITHUB_TOKEN }}
           allowed_non_write_users: '*'
+          allowed_bots: '*'
```
maskarb: I think the Amber review has a good point, allowing all is too broad. I think we ought to set this to the bots we know about first, like:
```
allowed_bots: "dependabot[bot],github-actions[bot]"
```
These are the only 2 I have seen with commits, so maybe we just stick to these and add more as we need to.

PR #819, opened 2026-03-05
perf(frontend): message windowing + stable keys in MessagesTab

Gkrumbach07:
## Summary

Addresses UI bogging-down and potential memory pressure caused by accumulating large numbers of AG-UI events in long-running sessions.

- **Message windowing**: cap the rendered DOM to the last 100 messages (`MAX_VISIBLE_MESSAGES`). A lightweight "Load earlier messages" button reveals older history on demand, keeping the number of mounted React subtrees bounded regardless of session length
- **Scroll position preservation**: `useLayoutEffect` + explicit `scrollTop` assignment anchors the viewport when loading earlier messages, overriding browser scroll-anchoring to get deterministic behavior
- **Stable React keys**: replace `key={sm-${idx}}` (index-based) with IDs derived from the message itself. Index keys cause React to re-render every message below an insertion point; stable keys mean only the changed node is reconciled
- **`handleScroll` guard**: `setIsAtBottom` now uses a functional update that short-circuits when the boolean value hasn't changed, eliminating the re-render triggered on every scroll event
- **`streamMessages` useMemo split**: the single O(n) memo that ran on every text delta is split into `committedStreamMessages` (hierarchy traversal, runs infrequently) and `streamMessages` (O(1) append, runs on every `TEXT_MESSAGE_CONTENT` event)

## Jira

https://issues.redhat.com/browse/RHOAIENG-52026

## Test plan

- [ ] Open a session with a large number of messages; confirm only the latest 100 are rendered (use DevTools Elements panel to count)
- [ ] Click "Load earlier messages"; confirm the viewport stays anchored to the same content and doesn't jump to the top
- [ ] Verify new streaming messages still auto-scroll when the user is at the bottom
- [ ] Verify the empty-state placeholder appears for sessions with no messages
- [ ] Confirm system-message toggle still works with the windowed list
- [ ] React DevTools Profiler: confirm `committedStreamMessages` does not re-run on `TEXT_MESSAGE_CONTENT` events

### DIFF
--- a/components/frontend/src/components/session/MessagesTab.tsx
+++ b/components/frontend/src/components/session/MessagesTab.tsx
@@ -4,6 +4,7 @@ import React, { useState, useRef, useEffect, useMemo, useLayoutEffect, useCallba
 import { MessageSquare } from "lucide-react";
 import { StreamMessage } from "@/components/ui/stream-message";
 import { LoadingDots } from "@/components/ui/message";
+import { Button } from "@/components/ui/button";
 import { ChatInputBox } from "@/components/chat/ChatInputBox";
 import { QueuedMessageBubble } from "@/components/chat/QueuedMessageBubble";
 import type { AgenticSession, MessageObject, ToolUseMessages, HierarchicalToolMessage } from "@/types/agentic-session";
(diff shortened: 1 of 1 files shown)


jeremyeder:
@Gkrumbach07 are you able to address the feedback so we can merge this?

Gkrumbach07:
Yeah I will. Afk rn will do it in a sec and merge it

PR #832, opened 2026-03-06
docs: add feature flag fail-modes reference and consolidate ADRs

maskarb:
- Add fail-modes.md documenting which evaluation methods fail open vs closed, where each is used, and the full evaluation precedence
- Move unleash ADR from components/backend/docs/adr/ to docs/internal/adr/ as ADR-0007 (0006 was already taken)
- Update ADR status from Proposed to Accepted, add sections for fail modes, model/runner feature gates, and flag sync at startup
- Fix incorrect "all flags disabled" claims in README and unleash guide to distinguish fail-closed (general) vs fail-open (model) behavior
- Update unleash guide with workspace override docs, full API endpoint list, Unleash server UI access instructions, and correct navigation path for the workspace feature flags panel
- Fix stale relative links in runner README and ADR


Gkrumbach07:
@mergifyio queue

PR #841, opened 2026-03-06
test "kustomize build" for overlays before merge

### LINKED ISSUE #840: add CI coverage for kustomize changes
```
```

Every pull request that touches `components/manifests/**` should validate that all kustomize overlays still build.

**Context:** Today only `production` and `e2e` get validated indirectly (via deploy workflows that run on push to `main`). `kind`, `kind-local`, and `local-dev` have no CI validation at all. A broken overlay is only discovered when someone runs `make kind-up` locally.

### Requirements

- [ ] Run `oc kustomize` for every overlay (`production`, `e2e`, `kind`, `kind-local`, `local-dev`) on PRs that change any file under `components/manifests/`
- [ ] Fail the check if any overlay fails to build

### Future: dry-run against stage cluster

A follow-up PR could add `oc apply --dry-run=server` against the stage cluster for the `production` overlay. This would validate admission webhooks and RBAC, not just kustomize syntax. Guard with `if: github.event.pull_request.head.repo.full_name == github.repository` to skip fork PRs where secrets are unavailable. Do not use `pull_request_target`.

### Out of scope

- Treating kustomize deprecation warnings (e.g. `commonLabels` → `labels`) as errors — track separately
- `mpp-preprod` overlay (RHOAIENG-51656) — does not exist on `main` yet; add to the overlay list when it lands

### Related

- #836 — kustomize component refactoring (this CI should land before or alongside that work)

ktdreyer:
Three of our five overlays (`kind`, `kind-local`, `local-dev`) had no CI validation. They can easily break silently if we change something in shared base resources.

Add basic syntax checking.

Closes #840

### DIFF
--- a/.github/workflows/kustomize-check.yml
+++ b/.github/workflows/kustomize-check.yml
@@ -1,0 +1,47 @@
+name: kustomize-build
+on:
+  pull_request:
+    paths:
+      - components/manifests/**
+      - .github/workflows/kustomize-check.yml
+  push:
+    paths:
+      - components/manifests/**
+      - .github/workflows/kustomize-check.yml
+
+  workflow_dispatch:
+
+concurrency:
+  group: kustomize-build-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  kustomize-build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Install oc
+        uses: redhat-actions/openshift-tools-installer@v1
+        with:
+          oc: latest
+
+      - name: Build all overlays
+        run: |
+          failed=0
+          for overlay in production e2e kind kind-local local-dev; do
+            echo "=== $overlay ==="
+            if oc kustomize "components/manifests/overlays/$overlay" > /dev/null; then
+              echo "OK"
+            else
+              echo "FAILED"
+              failed=1
+            fi
+            echo ""
+          done
+
+          if [ "$failed" -eq 1 ]; then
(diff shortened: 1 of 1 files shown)


ktdreyer:
I decided to go with a hard-coded overlay list because we should not be changing those without notice. Several things depend on them (ie make files, other GHA workflows, etc)

.github/workflows/kustomize-check.yml
```diff
@@ -0,0 +1,36 @@
...
+      - components/manifests/**
+
+jobs:
+  kustomize-build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install oc
+        uses: redhat-actions/oc-installer@v1
```
ktdreyer: The [README](https://github.com/redhat-actions/oc-installer?tab=readme-ov-file) indicates this is deprecated. https://github.com/redhat-actions/openshift-tools-installer is the replacement.

ktdreyer:
Addressing the Claude Code review items above:

> 1. `oc: latest` is unpinned

I did this by design to reduce maintenance.

> Hardcoded overlay list will silently miss future overlays

I did this by design. It's a safety check in case we **delete** overlays accidentally.

I did not want to dynamically discover overlays because I don't anticipate this section of the code changing *that* much over time.

> No explicit permissions block

The default is `read`. It's noisy and distracting to hard-code the default here.

> Double-run on PR branches

This is by design, so that I can test the action while I push to work-in-progress branches.

ktdreyer:
(rebased to latest `main` & squashed)

PR #846, opened 2026-03-08
fix(workflow): resolve stale greeting and race condition in startup prompt

jeremyeder:
## Summary

Addresses Critical and Major review feedback from code review of #780:

- **Critical — Stale greeting persists when switching workflows**: `setWorkflowGreeting` was only called when `startupPrompt` was truthy, so switching from a workflow with a greeting to one without left the old greeting displayed. Now clears unconditionally with `setWorkflowGreeting(workflow.startupPrompt || null)`.

- **Major — Race condition drops greeting silently**: The `useEffect` that activates queued workflows referenced `ootbWorkflows` but didn't include it in the dependency array. If the OOTB workflows API response hadn't resolved when the session reached "Running" phase, `fullWorkflow` would be `undefined` and `startupPrompt` silently lost. Now gates on `ootbWorkflows.length > 0` and includes `ootbWorkflows` in the dependency array.

## Test plan

- [ ] Select a workflow with a `startupPrompt` — greeting appears immediately
- [ ] Switch to a workflow without a `startupPrompt` — previous greeting is cleared
- [ ] Select a workflow before session reaches "Running" (queued activation) — greeting still appears after session starts
- [ ] Verify no duplicate workflow activations from the added dependency

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/components/frontend/package-lock.json
+++ b/components/frontend/package-lock.json
@@ -51,14 +51,32 @@
       "devDependencies": {
         "@eslint/eslintrc": "^3",
         "@tailwindcss/postcss": "^4",
+        "@testing-library/jest-dom": "^6.9.1",
+        "@testing-library/react": "^16.3.2",
         "@types/node": "^20",
         "@types/react": "^19",
         "@types/react-dom": "^19",
         "eslint": "^9",
         "eslint-config-next": "15.5.2",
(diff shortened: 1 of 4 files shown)


jeremyeder:
Closing as stale/superseded. Reopen if still needed.

PR #849, opened 2026-03-09
fix(workflow): render startup greeting client-side with race condition fixes

jeremyeder:
## Summary

Combines PR #780 and PR #846 into a single PR:

- **Backend**: Add `startupPrompt` field to `OOTBWorkflow` struct and parse it from `ambient.json` in workflow repos
- **Frontend**: Render workflow startup greeting client-side instead of round-tripping through Claude, fixing stale greeting and race conditions
- **Runner**: Remove server-side greeting logic (no longer needed with client-side approach)
- **Tests**: Add vitest regression tests for stale greeting bug in `use-workflow-management`

## What was broken

When a user clicked a workflow tile (e.g. "PRD RFE Workflow"), the workflow activated successfully but:
1. The `startupPrompt` from `ambient.json` was never served by the OOTB endpoint (missing field)
2. No greeting was displayed to guide the user on next steps
3. The session appeared to do nothing after activation

## Test plan

- [ ] Click a workflow tile on a Running session — verify greeting appears
- [ ] Switch between workflows — verify greeting updates (no stale state)
- [ ] Select workflow on a non-Running session — verify it queues and activates when Running
- [ ] Run `cd components/frontend && npx vitest run` — verify regression tests pass

Supersedes #780 and #846.

### DIFF
--- a/components/backend/handlers/sessions.go
+++ b/components/backend/handlers/sessions.go
@@ -1839,13 +1839,15 @@ func fetchGitHubDirectoryListing(ctx context.Context, owner, repo, ref, path, to
 
 // OOTBWorkflow represents an out-of-the-box workflow
 type OOTBWorkflow struct {
-	ID          string `json:"id"`
-	Name        string `json:"name"`
-	Description string `json:"description"`
-	GitURL      string `json:"gitUrl"`
-	Branch      string `json:"branch"`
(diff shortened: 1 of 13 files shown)


jeremyeder:
Closing as stale/superseded. Reopen if still needed.

PR #855, opened 2026-03-09
feat: Add session control endpoints to public API

tiwillia:
Enhances the public API to enable session control.

Prior to this change, the public API is limited to supporting CRUD on sessions. This lacked any ability to control and review session output. The following APIs are added as part of this change:

| Method | Path | Summary | Request Body | Success |
|--------|------|---------|--------------|---------|
| `POST` | `/sessions/{id}/message` | Send a message to a running session | `SendMessageRequest` | `202` |
| `GET` | `/sessions/{id}/output` | Get session output (transcript, compact, or raw events) | — | `200` |
| `POST` | `/sessions/{id}/runs` | Create an AG-UI run (full control over run input) | `CreateRunRequest` | `202` |
| `GET` | `/sessions/{id}/runs` | List all runs in a session | — | `200` |
| `POST` | `/sessions/{id}/start` | Resume a stopped/completed session | — | `202` |
| `POST` | `/sessions/{id}/stop` | Stop a running session (terminates pod) | — | `202` |
| `POST` | `/sessions/{id}/interrupt` | Interrupt current run without killing the session | — | `200` |

These endpoints make it possible to _orchestrate_ sessions using the public API. This can enable external tools to automate session orchestration and, with RBAC changes, could enable session agents to interact with other sessions through gathering output and sending messages.

This change also contains several fixes against the existing session CRUD endpoints, including:
- Adding `display_name` support
- Fixing how repos and the initial prompt are communicated to the backend (these are just dropped in the current implementation).
- Fixing an issue where large request body output (GET /sessions) were cancelled before returned

**Note** that the public API is currently _unused_ by any existing client, UI, etc - but is proposed to be the primary way clients interact with the platform. This change assumes that is still the current intent, as all of this functionality is already available in the backend API.

### DIFF
--- a/components/public-api/handlers/sessions.go
+++ b/components/public-api/handlers/sessions.go
@@ -147,6 +147,21 @@ func CreateSession(c *gin.Context) {
 		}
 		backendReq["repos"] = repos
 	}
+	if req.ActiveWorkflow != nil && strings.TrimSpace(req.ActiveWorkflow.GitURL) != "" {
+		wf := map[string]interface{}{
+			"gitUrl": req.ActiveWorkflow.GitURL,
+		}
+		if req.ActiveWorkflow.Branch != "" {
+			wf["branch"] = req.ActiveWorkflow.Branch
+		}
+		if req.ActiveWorkflow.Path != "" {
+			wf["path"] = req.ActiveWorkflow.Path
+		}
+		backendReq["activeWorkflow"] = wf
+	}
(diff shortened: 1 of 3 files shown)


jeremyeder:
@tiwillia is this still needed?

tiwillia:
@jeremyeder from what I heard, the public API is considered deprecated at this point and won't actually be replacing the backend API as suggested in the proposal docs in the repo that prompted me to push for this change.

So I'll close this.

PR #859, opened 2026-03-09
fix(RHOAIENG-52390): add kubectl context safety check

mprpic:
Adds a check-local-context prerequisite that verifies the current kubectl context points to a local cluster (kind- or minikube) before running targets that modify cluster state. Prevents accidentally applying manifests or restarting deployments on a remote cluster.

Guarded targets: kind-rebuild, kind-port-forward, local-down, local-rebuild, local-reload-backend, local-reload-frontend, local-reload-operator, dev-bootstrap.

Bypassable with SKIP_CONTEXT_CHECK=true for edge cases.


ktdreyer:
@mergifyio queue

ktdreyer:
(trying manually putting this into the queue, ahead of #872 )

PR #870, opened 2026-03-10
chore(ci): remove stale public-api from build and deploy workflows

maskarb:
## Summary

- Remove all `public-api` references from `components-build-deploy.yml` and `prod-release-deploy.yaml`
- The public-api component is no longer used; its presence in the workflows was stale

## Test plan

- [ ] Verify workflow syntax is valid (GitHub Actions lint)
- [ ] Confirm no other workflows reference public-api

### DIFF
--- a/.github/workflows/components-build-deploy.yml
+++ b/.github/workflows/components-build-deploy.yml
@@ -10,7 +10,6 @@ on:
       - 'components/operator/**'
       - 'components/backend/**'
       - 'components/frontend/**'
-      - 'components/public-api/**'
       - 'components/ambient-api-server/**'
   pull_request:
     branches: [main]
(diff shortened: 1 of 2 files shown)


markturansky:
hold until the new API is fully functional

jeremyeder:
@Gkrumbach07 @markturansky — is public-api still needed? The component directory and CI pipelines still reference it. If we're decommissioning, we should remove the source + manifests too, not just the CI references. If it's still in use, we should close this PR. Please advise.

Gkrumbach07:
@jeremyeder only the MCP currently references it, but it's very easy for us to go back to just pointing out the back end API for now instead of the public API.

PR #878, opened 2026-03-10
feat(frontend): add dynamic page titles for workspace and session pages

mprpic:
Replaces https://github.com/ambient-code/platform/pull/861/

### DIFF
--- a/components/frontend/src/app/projects/[name]/sessions/[sessionName]/layout.tsx
+++ b/components/frontend/src/app/projects/[name]/sessions/[sessionName]/layout.tsx
@@ -1,0 +1,7 @@
+export default function SessionLayout({
+  children,
+}: {
+  children: React.ReactNode;
+}) {
+  return children;
+}
--- a/components/frontend/src/app/projects/[name]/sessions/[sessionName]/page.tsx
+++ b/components/frontend/src/app/projects/[name]/sessions/[sessionName]/page.tsx
@@ -251,6 +251,14 @@ export default function ProjectSessionDetailPage({
     return undefined;
   }, [capabilities?.framework, runnerTypes]);
 
+  // Update page title with display name when available
+  useEffect(() => {
+    const title = session?.spec?.displayName || sessionName;
+    if (title) {
+      document.title = `${title} · Ambient Code Platform`;
+    }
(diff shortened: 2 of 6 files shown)


components/frontend/src/app/projects/[name]/sessions/[sessionName]/layout.tsx
```diff
@@ -0,0 +1,18 @@
+import type { Metadata } from "next";
+
+export async function generateMetadata({
+  params,
+}: {
+  params: Promise<{ sessionName: string }>;
+}): Promise<Metadata> {
+  const { sessionName } = await params;
+  return { title: `${sessionName} · Ambient Code Platform` };
```
Comment (christianvogt) [inline components/frontend/src/app/projects/[name]/sessions/[sessionName]/layout.tsx]: @mprpic this is issue I want to avoid. The name of the tab ends up being useless because it looks like `session-17734238534 · Ambient Code Platform`

I want this to be the display name of the session.

Comment (mprpic) [reply components/frontend/src/app/projects/[name]/sessions/[sessionName]/layout.tsx]: Ah, I missed that in your original PR. Before this though, it didn't contain any dynamic content and was just `Ambient Code Platform`.

Anyway, latest commit fixes it to be the display name first.

image

components/frontend/src/app/projects/[name]/sessions/[sessionName]/page.tsx
```diff
@@ -251,6 +251,14 @@ export default function ProjectSessionDetailPage({
     return undefined;
   }, [capabilities?.framework, runnerTypes]);
 
+  // Update page title with display name when available
+  useEffect(() => {
+    const title = session?.spec?.displayName || sessionName;
+    if (title) {
+      document.title = `${title} · Ambient Code Platform`;
+    }
+  }, [session?.spec?.displayName, sessionName]);
```
Comment (christianvogt) [inline components/frontend/src/app/projects/[name]/sessions/[sessionName]/page.tsx:260]: If the next.js way is to use metadata, why not just make the request to obtain the session name in the metadata function?

What I originally did was use the react `title` component which I think is a better approach vs a `useEffect`. While it may not make a difference here, if nested components want to add a title, on unmount it would restore the parent value.

Comment (mprpic) [reply components/frontend/src/app/projects/[name]/sessions/[sessionName]/page.tsx:260]: Ack, let's go with your PR. Closing!

PR #895, opened 2026-03-12
fix(RHOAIENG-52949): scope session listing by X-Ambient-Project header

maskarb:
## Summary

The session list endpoint (`GET /api/ambient/v1/sessions`) returned all sessions from the database regardless of the `X-Ambient-Project` header, breaking multi-tenant isolation.

## Root Cause

The handler already supported `project_id` as a query parameter filter, but the SDK and CLI send the project via the `X-Ambient-Project` header — which was not read by the list handler.

## Fix

Read `X-Ambient-Project` header as fallback when the `?project_id` query param is not set. The same validated `project_id = 'X'` filter is applied to scope results. Query param takes precedence over header if both are set.

## Test plan

- [ ] `acpctl get sessions` with project set to `A` should only return sessions from project `A`
- [ ] `GET /sessions?project_id=A` still works (query param precedence)
- [ ] `GET /sessions` without header or param returns all sessions (backward compatible)

**Jira:** [RHOAIENG-52949](https://issues.redhat.com/browse/RHOAIENG-52949)

### DIFF
--- a/components/ambient-api-server/plugins/common/project_scope_test.go
+++ b/components/ambient-api-server/plugins/common/project_scope_test.go
@@ -1,0 +1,151 @@
+package common
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/openshift-online/rh-trex-ai/pkg/services"
+)
+
+func newRequest(queryParams, headerProject string) *http.Request {
+	reqURL := "/sessions"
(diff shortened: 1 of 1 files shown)


Gkrumbach07:
@coderabbitai full review

Gkrumbach07:
@coderabbitai full review

Gkrumbach07:
@mergifyio queue

PR #927, opened 2026-03-16
fix(frontend): handle simple question format in AskUserQuestion HITL UI

Gkrumbach07:
## Summary

- Fixes bug where the HITL (Human-in-the-Loop) user input UI never appeared when the agent called `AskUserQuestion`
- The component only handled structured `{ questions: [...] }` input, but Claude Code sends a simple `{ question: "..." }` string format — causing `parseQuestions` to return `[]` and the component to render nothing
- Adds fallback parsing for the simple format and a free-form text input when no predefined options exist

## Test plan

- [ ] Trigger an `AskUserQuestion` tool call in a session (e.g. ask the agent something that requires clarification)
- [ ] Verify the HITL question UI renders with a text input field
- [ ] Verify submitting an answer works and the form disables after submission
- [ ] Verify the structured `{ questions: [...] }` format still works (radio buttons, checkboxes)
- [ ] `npx vitest run` — all 471 tests pass

### DIFF
--- a/components/frontend/src/components/session/ask-user-question.tsx
+++ b/components/frontend/src/components/session/ask-user-question.tsx
@@ -31,6 +31,13 @@ function parseQuestions(input: Record<string, unknown>): AskUserQuestionItem[] {
   if (isAskUserQuestionInput(input)) {
     return input.questions;
   }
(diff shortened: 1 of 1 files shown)


Gkrumbach07:
@mergifyio queue

PR #929, opened 2026-03-16
fix: flaky "element detached from DOM" in sessions test

mprpic:
Wait for the sessions list API response before clicking the new session button, ensuring React has finished re-rendering with the fetched data.

Example failed test:
https://github.com/ambient-code/platform/actions/runs/23146894542/job/67237433020


Gkrumbach07:
@mergifyio queue

Issue #940, opened 2026-03-17
coderabbit reviews should not gate mergify auto-merges


ktdreyer:
CodeRabbit does not always leave a "review" on PRs. Example: https://github.com/ambient-code/platform/pull/933

We should drop these criteria lines from `.mergify.yml`:

```
commented-reviews-by=coderabbitai[bot]
```

PR #941, opened 2026-03-17
mergify: remove coderabbit comment requirement

ktdreyer:
CodeRabbit does not leave a review if there are no actionable comments.  When Mergify does not see the CodeRabbit review, it will not merge the PR, even if the PR meets all the other criteria (successful human review + passing tests).

The purpose of adding coderabbit to the Mergify criteria is to provide a natural quality governer: to ensure that humans read the reviews. Turns out CodeRabbit is useful enough that we don't need to enforce this with Mergify, and tweaking it further is not worth optimizing.

Fixes: #940

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -6,7 +6,6 @@ queue_rules:
       - "#approved-reviews-by>=2"
       - "#changes-requested-reviews-by=0"
       - check-success=End-to-End Tests
-      - commented-reviews-by=coderabbitai[bot]
       - check-success=validate-manifests
       - check-success=test-local-dev-simulation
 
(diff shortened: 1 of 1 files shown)


jeremyeder:
Hey, something to think about - we will need to find a type of PR that we
are ok auto merging.  Are there stats collected from mergify that we could
learn from?


-- Jeremy Eder (he/him/his)

.mergify.yml
```diff
@@ -6,7 +6,6 @@ queue_rules:
       - "#approved-reviews-by>=2"
       - "#changes-requested-reviews-by=0"
       - check-success=End-to-End Tests
-      - commented-reviews-by=coderabbitai[bot]
```
mprpic, line 9: It's too bad there isn't a generic `commented-on-by` that counts _any_ comment, not just review comments.

ktdreyer:
@mergifyio queue

PR #965, opened 2026-03-19
feat(runner): add backend API tools for session management

Gkrumbach07:
## Summary

Add custom Claude Agent SDK tools to enable running sessions to interact with the backend API until the MCP server is ready. This provides a temporary solution for session management operations **and API integration knowledge**.

### New Components

- **BackendAPIClient**: HTTP client for backend API calls with BOT_TOKEN authentication
- **Backend MCP tools (acp server)**: **6 tools** for session management and API learning
- **Comprehensive unit tests**: 20 tests covering all functionality

### Tools Provided

The tools are automatically registered in the ClaudeBridge MCP configuration when `BACKEND_API_URL` and `PROJECT_NAME` environment variables are set.

#### Action Tools (DO things)
- `acp_list_sessions`: List active sessions in the current project
- `acp_get_session`: Get detailed session information
- `acp_create_session`: Create new session with repos/prompts/model
- `acp_stop_session`: Stop a running session
- `acp_send_message`: Send message to session (async)

#### Knowledge Tool (TEACH the agent)
- `acp_get_api_reference`: Returns comprehensive API documentation so agents can generate code (HTML dashboards, scripts, etc.) that calls our backend API directly

### API Integration Use Case

**User**: "Create an HTML dashboard with a button that creates a new session"

**Agent workflow**:
1. Calls `acp_get_api_reference()` to learn the API
2. Gets complete docs: endpoints, auth, payloads, code examples
3. Generates working HTML with proper `fetch()` calls

The API reference tool returns:
- All endpoints (methods, paths, request/response formats)
- Authentication patterns (Bearer token, proxy examples)
- Working code examples (JavaScript, Python, HTML)
- Session name validation rules
- Error handling patterns
- Full integration examples

### API Integration

The tools use the existing backend REST API:
- `GET /api/projects/:projectName/agentic-sessions` - list sessions
- `GET /api/projects/:projectName/agentic-sessions/:sessionName` - get session
- `POST /api/projects/:projectName/agentic-sessions` - create session
- `POST /api/projects/:projectName/agentic-sessions/:sessionName/stop` - stop session
- `POST /api/projects/:projectName/agentic-sessions/:sessionName/agui/run` - send message

### Test Plan

- [x] All unit tests pass (20 new tests + 472 existing tests)
- [x] Code passes ruff linting and formatting
- [ ] Manual testing with a live session (requires deployment)
- [ ] Verify tools appear in Claude Code's tool list
- [ ] Test each tool operation against running backend
[...]

PR #962, opened 2026-03-19
fix(frontend): fetch models from API instead of using hardcoded lists

Gkrumbach07:
## Summary
- Removed hardcoded model lists from `runner-model-selector` component
- Replaced with API calls to `GET /api/projects/:projectName/models?provider=X`
- Updated `new-session-view` to fetch default model from API
- Fixed tests to mock API instead of testing hardcoded data

## Problem
The new session UI (`runner-model-selector.tsx`) was using hardcoded model lists that only included 4.5 models:
```typescript
const MODELS_BY_RUNNER = {
  "claude-code": [
    { id: "claude-haiku-4-5", ... },
    { id: "claude-sonnet-4-5", ... },
    { id: "claude-opus-4-5", ... },  // Only 4.5!
  ],
  ...
};
```

This caused **Claude 4.6 models to be missing** in the new session view, even though:
- The backend API properly returns all models (including 4.6)
- The `models.json` ConfigMap includes 4.6 models
- The older `create-session-dialog` component fetches from API correctly

## Root Cause
Recent UI redesign introduced `runner-model-selector` with hardcoded models instead of fetching from the API. The old `create-session-dialog` was already using the API correctly via `useModels()` hook.

## Solution
Updated `runner-model-selector` to use the same pattern as `create-session-dialog`:
1. Import and call `useModels(projectName, enabled, provider)` hook
2. Fetch models filtered by runner's provider from API
3. Map API response to select options
4. Remove all hardcoded model constants

### `runner-model-selector.tsx`
- Added `useModels()` import
- Removed `MODELS_BY_RUNNER` constant (hardcoded)
- Removed `getModelsForRunner()` function
- Updated `getDefaultModel()` to use API models list
- Added `RunnerModelsRadioGroup` component to fetch models per runner
- Updated exports to remove deleted functions

### `new-session-view.tsx`
- Added `useModels()` import
- Fetch models when component mounts
- Use API's `defaultModel` instead of calling `getDefaultModel(runnerId)`
- Update default model when modelsData changes

### Tests
- Updated mocks to provide `useModels` data
- Changed test assertions to match new API-driven behavior
- Removed tests for deleted `getModelsForRunner()` function
- Updated `getDefaultModel` tests for new signature

## Test plan
- [ ] Verify 4.6 models appear in new session view model selector
- [ ] Verify models update when switching runners
- [ ] Verify default model is selected correctly
- [ ] Frontend unit tests pass
- [ ] TypeScript builds with 0 errors

[...]

PR #971, opened 2026-03-19
fix(e2e): add visibility waits for new-session-btn to fix timing issues

Gkrumbach07:
## Summary

Fixes the E2E test failures across multiple PRs by adding proper visibility checks before clicking the `new-session-btn`.

## Problem

The E2E tests were failing with:
```
AssertionError: Timed out retrying after 10000ms: Expected to find element: `[data-testid="new-session-btn"]`, but never found it.
```

This caused all 58 tests to be skipped in the before() hook, failing the E2E suite on PRs #962, #964, #965, and others.

## Root Cause

The workspace page (`/projects/{slug}`) shows a loading spinner while fetching project data. The test's `cy.request()` API calls complete successfully, but the UI is still rendering. When the test immediately tries to click `new-session-btn`, the page is still showing the loading state, so the button doesn't exist yet.

## Solution

Added `should('be.visible')` checks before clicking the button in three places:
1. `before()` hook (line 80) - with 20s timeout
2. "should open create session dialog" test (line 184)  
3. "Agent Interaction" test (line 248)

This ensures the button is fully rendered and interactive before the test attempts to click it.

## Test Plan

- [x] Local syntax validation (eslint passed)
- [x] Committed and pushed to branch
- [ ] CI E2E tests should now pass (verifying...)
- [ ] Should fix failures on PRs #962, #964, #965

### DIFF
--- a/e2e/cypress/e2e/sessions.cy.ts
+++ b/e2e/cypress/e2e/sessions.cy.ts
@@ -77,7 +77,8 @@ describe('Ambient Session Management Tests', () => {
       body: { data: { ANTHROPIC_API_KEY: apiKey } }
     })).then((r) => expect(r.status).to.eq(200))
 
-    // Create a session for UI tests
+    // Wait for the workspace page to finish loading before creating a session
+    cy.get('[data-testid="new-session-btn"]', { timeout: 20000 }).should('be.visible')
     cy.get('[data-testid="new-session-btn"]').click()
     cy.get('[data-testid="create-session-submit"]', { timeout: 10000 })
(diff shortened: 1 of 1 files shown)


Gkrumbach07:
Closing this PR as obsolete. The E2E test timing issues have already been fixed on main via commit da4285c9 and commit 6a487e00, which refactored session creation to use direct API calls instead of UI interactions. This is a more robust solution than adding visibility waits, as it completely bypasses the UI rendering timing issues. The failures seen on PRs #962, #964, and #965 were on an older version of main that has since been updated.

PR #991, opened 2026-03-23
fix(frontend): resolve agent response buffering due to stale EventSou…

psav:
…rce closure

Fixes intermittent message buffering where agent responses don't display until the user sends another message. The EventSource onmessage handler was capturing a stale processEvent closure, causing events to be processed with outdated setState callbacks when component callbacks changed (e.g., during agent team operations). Use a ref to ensure the handler always calls the latest processEvent, guaranteeing immediate UI updates.

### DIFF
--- a/components/frontend/src/hooks/use-agui-stream.ts
+++ b/components/frontend/src/hooks/use-agui-stream.ts
@@ -185,7 +185,7 @@ export function useAGUIStream(options: UseAGUIStreamOptions): UseAGUIStreamRetur
         }, delay)
       }
     },
-    [projectName, sessionName, processEvent, onConnected, onError, onDisconnected],
+    [projectName, sessionName, onConnected, onError, onDisconnected],
   )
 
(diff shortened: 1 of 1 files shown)


Gkrumbach07:
@psav can u look at the code rabbit comment. is it accurate 

> Remove processEvent from connect's dependency array.

> Since the handler now accesses processEvent through the ref rather than capturing it directly, processEvent should be removed from the dependency array at line 188. Keeping it causes unnecessary recreations of connect whenever processEvent changes, which defeats part of the purpose of using the ref pattern.****

jeremyeder:
I rechecked the current head, and these points are already addressed:

- `processEventRef` is synchronized with `useLayoutEffect`
- `processEvent` has been removed from `connect`’s dependency array
- there is a regression test covering rerender without reconnecting and asserting the latest callback is used

So my earlier concern was accurate for an earlier revision, but it’s stale against the current PR state.

PR #997, opened 2026-03-24
fix(runner): propagate userId to all Langfuse traces across async requests

jeremyeder:
## Summary

- ~77% of Langfuse traces had no `userId`, `sessionId`, or `tags`, making usage reports unreliable
- Root cause: `propagate_attributes()` stores values in OTel contextvars, which are per-asyncio-task. The context set during `initialize()` (first HTTP request) was invisible to subsequent requests running in new asyncio tasks
- Fix: save propagation args during `initialize()` and re-enter `propagate_attributes()` at the start of each turn in `start_turn()`. Per-turn context is cleaned up in `end_turn()`, `finalize()`, and error paths

## Test plan

- [x] 474 runner unit tests pass
- [ ] Deploy to dev cluster and verify new traces in Langfuse have `userId` and `sessionId` populated
- [ ] Run `scripts/langfuse-usage-report.py` and confirm "unknown" bucket is significantly reduced

### DIFF
--- a/components/runners/ambient-runner/ambient_runner/observability.py
+++ b/components/runners/ambient-runner/ambient_runner/observability.py
@@ -462,6 +462,8 @@ class ObservabilityManager:
             )
 
         except Exception as e:
+            self._current_turn_generation = None
+            self._current_turn_ctx = None
             self._exit_turn_propagate_ctx()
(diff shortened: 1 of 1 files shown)


jeremyeder:
example

image

PR #1007, opened 2026-03-24
Fix workspace name uniqueness to be scoped per user

derekhiggins:
Previously workspace names were globally unique across all users, preventing users from creating workspaces with names already taken by others (e.g., 'ambient-code').

Changes:
- Add owner_user_id field to Project model
- Change unique constraint from name-only to composite (owner_user_id, name)
- Automatically set owner_user_id from authenticated user context
- Add migration to update existing database schema
- Update OpenAPI schema to include owner_user_id field
- Fix ID generation to use unique IDs instead of name

This allows each user to have their own workspaces with any name, with uniqueness scoped only within their own workspace collection.

--
Done in ambient-code, Needs review (I got no knowledge of this codebase)

### DIFF
--- a/components/ambient-api-server/openapi/openapi.projects.yaml
+++ b/components/ambient-api-server/openapi/openapi.projects.yaml
@@ -215,6 +215,10 @@ components:
           required:
             - name
           properties:
+            owner_user_id:
+              type: string
+              description: User ID of the project owner (set automatically by the server)
+              readOnly: true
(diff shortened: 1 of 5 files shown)


jeremyeder:
API is a work in progress, so let's hold this until it rolls out. Imminent.

derekhiggins:
Closing as not a current priority

Issue #952, opened 2026-03-18
markdown docs are consistently out-of-date


ktdreyer:
As humans and agents read through the docs, several READMEs or other markdown files are consistently wrong or out-of-date. This can steer the LLMs into amplifying bad patterns. (garbage in, garbage out). 

Examples I remember:

https://github.com/ambient-code/platform/issues/790
https://github.com/ambient-code/platform/issues/835
https://github.com/ambient-code/platform/issues/769

This is somewhat related to code churn: it's low value to write a README for areas of the codebase that are high-churn (example: listing directory contents). It's much higher value to document low-churn, higher-level principles that change slowly (if at all).

Two approaches here:
1. "stop the bleeding" one-time pass to fix all the issues (see [this skill](https://github.com/ktdreyer/dotfiles/commit/18545f6805143851cd9db9c2257485dc7496f8d4) as inspiration)
2. Edit CLAUDE.md to stop Claude from writing brittle things into individual README Markdown files, and treat the code as authoritative in areas where it makes sense (ie, don't AI-generate a list of files in a directory into a Markdown file).

jeremyeder:
Housekeeping: reviewing open issues. @ktdreyer is this still relevant / something you plan to work on? If not, we'll close it out. No pressure either way — just trimming the backlog.

jeremyeder:
Hey there! 👋 Closing this out as part of a triage pass — it's been idle for a while and we want to keep the backlog focused. If this is still relevant or you'd like to pick it back up, feel free to reopen anytime. Thanks!

PR #793, opened 2026-03-04
RHOAIENG-50709: add ADR for CI/CD on firewalled clusters

ktdreyer:
Present a phased approach for running GitHub Actions inside firewalled OpenShift.

Phase 1 deploys a standalone runner with no CRDs or cluster-level permissions.

Phase 2 upgrades to Actions Runner Controller (ARC) if IT approves CRDs.

https://issues.redhat.com/browse/RHOAIENG-50709

Issue #768, opened 2026-03-02
deploy.sh requires standalone kustomize binary


ktdreyer:
**Jira:** [RHOAIENG-51905](https://issues.redhat.com/browse/RHOAIENG-51905)

## Problem

`components/manifests/deploy.sh` requires a standalone `kustomize` binary. It checks for it with `command_exists kustomize` and calls it directly throughout the script (`kustomize build`, `kustomize edit set image`, `kustomize edit set namespace`, etc.).

This is unnecessary because `kubectl` already ships with `kubectl kustomize` built in, and `kubectl` is already a prerequisite for this script.

## Proposal

Replace all `kustomize` calls with `kubectl kustomize` and remove the standalone binary check.

Note: the e2e deploy script (`e2e/scripts/deploy.sh`) already uses `kubectl kustomize` and does not have this problem.

_(Written by Claude Code / Opus 4.6)_

ktdreyer:
## Update: `kustomize edit` is not available via `kubectl`

Research shows that `kubectl kustomize` only integrates the **build** functionality from standalone kustomize — the `edit` subcommand was intentionally excluded from kubectl ([KEP discussion](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cli/2377-Kustomize/README.md), [PR #73033](https://github.com/kubernetes/kubernetes/pull/73033)).

This script uses both `kustomize build` and `kustomize edit set image`/`kustomize edit set namespace`. Replacing `kustomize build` with `kubectl kustomize` is straightforward, but we would need to choose a solution for the `edit` calls:
- Replace with `sed`
- Evaluate each `edit` call to see if we could simplify it to use fully-declarative configurations

jeremyeder:
Housekeeping: reviewing open issues. @ktdreyer is this still relevant / something you plan to work on? If not, we'll close it out. No pressure either way — just trimming the backlog.

jeremyeder:
Hey there! 👋 Closing this out as part of a triage pass — it's been idle for a while and we want to keep the backlog focused. If this is still relevant or you'd like to pick it back up, feel free to reopen anytime. Thanks!

PR #1032, opened 2026-03-25
feat(ci): ephemeral PR test instances on MPP dev cluster

markturansky:
## Summary

- **Build workflow**: All 7 component images are now built and pushed to quay on every PR (`pr-<N>-amd64` tag), not just changed components
- **Namespace provisioner** (`components/pr-test/provision.sh`): Creates/destroys `TenantNamespace` CRs on `ambient-code--config`; capacity-gated at 5 concurrent S0.x instances
- **Ambient installer** (`components/pr-test/install.sh`): Deploys production manifests with PR image tags; copies `ambient-vertex` and `ambient-api-server` secrets from source namespace via `oc`; mirrors Vertex AI config from `operator-config`
- **GitHub workflows**: `pr-e2e-openshift.yml` (provision → install → e2e → teardown on build completion) and `pr-namespace-cleanup.yml` (safety-net teardown on PR close)
- **Skills**: `.claude/skills/ambient/` (install on any OpenShift namespace) and `.claude/skills/ambient-pr-test/` (full PR test workflow)
- **Developer guide**: `openshift.md` trimmed to internal-registry-only content with links to skills

## Test plan

- [ ] Add `TEST_OPENSHIFT_SERVER` and `TEST_OPENSHIFT_TOKEN` GitHub Actions secrets pointing at `dev-spoke-aws-us-east-1`
- [ ] Verify build workflow pushes images tagged `pr-<N>-amd64` to quay on this PR
- [ ] Verify `pr-e2e-openshift.yml` triggers and provisions `ambient-code--pr-<N>-*` namespace
- [ ] Verify Ambient deploys and health check passes
- [ ] Verify E2E tests run against the ephemeral instance
- [ ] Verify namespace is torn down after workflow completes
- [ ] Verify `pr-namespace-cleanup.yml` fires on PR close

### DIFF
--- a/.github/workflows/components-build-deploy.yml
+++ b/.github/workflows/components-build-deploy.yml
@@ -120,8 +120,11 @@ jobs:
           elif [ -n "$SELECTED" ]; then
             # Dispatch with specific components
             FILTERED=$(echo "$ALL_COMPONENTS" | jq -c --arg sel "$SELECTED" '[.[] | select(.name as $n | $sel | split(",") | map(gsub("^\\s+|\\s+$";"")) | index($n))]')
+          elif [ "$EVENT" == "pull_request" ]; then
+            # PR — always build all components so every PR image is in quay for deployment
+            FILTERED="$ALL_COMPONENTS"
           else
-            # Push or PR — only changed components
+            # Push to main — only changed components
             FILTERED="[]"
             for comp in $(echo "$ALL_COMPONENTS" | jq -r '.[].name'); do
               if [ "${FILTER_MAP[$comp]}" == "true" ]; then
@@ -200,17 +203,18 @@ jobs:
           cache-from: type=gha,scope=${{ matrix.component.name }}-${{ matrix.arch.suffix }}
           cache-to: type=gha,mode=max,scope=${{ matrix.component.name }}-${{ matrix.arch.suffix }}
 
-      - name: Build ${{ matrix.component.name }} (${{ matrix.arch.suffix }}) for pull request
+      - name: Build and push ${{ matrix.component.name }} (${{ matrix.arch.suffix }}) for pull request
         if: github.event_name == 'pull_request'
         uses: docker/build-push-action@v6
         with:
           context: ${{ matrix.component.context }}
           file: ${{ matrix.component.dockerfile }}
           platforms: ${{ matrix.arch.platform }}
-          push: false
+          push: true
           tags: ${{ matrix.component.image }}:pr-${{ github.event.pull_request.number }}-${{ matrix.arch.suffix }}
           build-args: AMBIENT_VERSION=${{ github.sha }}
           cache-from: type=gha,scope=${{ matrix.component.name }}-${{ matrix.arch.suffix }}
+          cache-to: type=gha,mode=max,scope=${{ matrix.component.name }}-${{ matrix.arch.suffix }}
 
   merge-manifests:
     needs: [detect-changes, build]
--- a/.github/workflows/pr-e2e-openshift.yml
+++ b/.github/workflows/pr-e2e-openshift.yml
@@ -1,0 +1,183 @@
+name: PR E2E on OpenShift
+
+on:
+  workflow_run:
+    workflows: ["Build and Push Component Docker Images"]
+    types: [completed]
+
+concurrency:
+  group: pr-e2e-openshift-${{ github.event.workflow_run.pull_requests[0].number }}
+  cancel-in-progress: true
+
+jobs:
+  setup:
+    if: >
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success' &&
+      github.event.workflow_run.pull_requests[0] != null
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    outputs:
+      pr_number: ${{ steps.ctx.outputs.pr_number }}
+      instance_id: ${{ steps.ctx.outputs.instance_id }}
+      namespace: ${{ steps.ctx.outputs.namespace }}
+      image_tag: ${{ steps.ctx.outputs.image_tag }}
+    steps:
+      - name: Derive PR context
+        id: ctx
+        env:
+          PR_NUMBER: ${{ github.event.workflow_run.pull_requests[0].number }}
+          HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
+        run: |
+          # Sanitize branch name — allow only alphanumeric and separators
+          SAFE_BRANCH=$(echo "$HEAD_BRANCH" | tr '[:upper:]' '[:lower:]' \
+            | sed 's/[^a-z0-9]/-/g' | sed 's/-\+/-/g' | sed 's/^-\|-$//g' | cut -c1-64)
+
+          # Budget: "ambient-code--" (14) + "pr-${PR_NUMBER}-" (4+digits) + slug
+          # Max namespace = 63. Slug budget = 63 - 14 - 4 - ${#PR_NUMBER}
+          PR_LEN=${#PR_NUMBER}
+          SLUG_MAX=$(( 63 - 14 - 4 - PR_LEN ))
+          BRANCH_SLUG="${SAFE_BRANCH:0:$SLUG_MAX}"
+
+          INSTANCE_ID="pr-${PR_NUMBER}-${BRANCH_SLUG}"
+          NAMESPACE="ambient-code--${INSTANCE_ID}"
+          IMAGE_TAG="pr-${PR_NUMBER}-amd64"
+
+          echo "pr_number=${PR_NUMBER}"     >> $GITHUB_OUTPUT
+          echo "instance_id=${INSTANCE_ID}" >> $GITHUB_OUTPUT
+          echo "namespace=${NAMESPACE}"     >> $GITHUB_OUTPUT
+          echo "image_tag=${IMAGE_TAG}"     >> $GITHUB_OUTPUT
+
+          echo "PR:        ${PR_NUMBER}"
+          echo "Instance:  ${INSTANCE_ID}"
+          echo "Namespace: ${NAMESPACE}"
+          echo "Image tag: ${IMAGE_TAG}"
+          echo "NS length: ${#NAMESPACE}"
+
+  provision:
+    needs: setup
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout main (trusted scripts only)
+        uses: actions/checkout@v6
+        with:
+          ref: main
+
+      - name: Install oc
+        uses: redhat-actions/oc-installer@v1
+        with:
+          oc_version: 'latest'
+
+      - name: Log in to OpenShift
+        run: |
+          oc login "${{ secrets.TEST_OPENSHIFT_SERVER }}" \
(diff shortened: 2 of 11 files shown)


.github/workflows/pr-e2e-openshift.yml
```diff
@@ -0,0 +1,187 @@
...
+          ref: ${{ github.event.workflow_run.head_sha }}
+
+      - name: Install oc
+        uses: redhat-actions/oc-installer@v1
+        with:
+          oc_version: 'latest'
+
+      - name: Install kustomize
+        run: |
+          curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash
```
ktdreyer: We install `kustomize` in a few places (eg #768). We should stop doing that and use `oc kustomize` instead.

.github/workflows/pr-e2e-openshift.yml
```diff
@@ -0,0 +1,187 @@
...
+          sudo mv kustomize /usr/local/bin/
+
+      - name: Log in to OpenShift
+        run: |
+          oc login ${{ secrets.TEST_OPENSHIFT_SERVER }} \
+            --token=${{ secrets.TEST_OPENSHIFT_TOKEN }} \
+            --insecure-skip-tls-verify
```
ktdreyer: I realize we're copying `--insecure-skip-tls-verify` from other GH Actions, but do we need it? I thought ROSA obviates this.

ktdreyer:
TODO: Let's identify if/where this runs untrusted code and stamp that out. GH gives us some clues in the PR review here. I also filed #1034 to add another tool in the box.

.github/workflows/pr-e2e-openshift.yml
markturansky: good catch. this is from testing and should not be in any full deployment

.github/workflows/components-build-deploy.yml
```diff
@@ -200,17 +203,18 @@ jobs:
...
         with:
           context: ${{ matrix.component.context }}
           file: ${{ matrix.component.dockerfile }}
           platforms: ${{ matrix.arch.platform }}
-          push: false
+          push: true
           tags: ${{ matrix.component.image }}:pr-${{ github.event.pull_request.number }}-${{ matrix.arch.suffix }}
           build-args: AMBIENT_VERSION=${{ github.sha }}
           cache-from: type=gha,scope=${{ matrix.component.name }}-${{ matrix.arch.suffix }}
+          cache-to: type=gha,mode=max,scope=${{ matrix.component.name }}-${{ matrix.arch.suffix }}
```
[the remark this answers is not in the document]
ktdreyer: I agree with this

.github/workflows/pr-e2e-openshift.yml
```diff
@@ -0,0 +1,183 @@
...
+          ref: main
+
+      - name: Install oc
+        uses: redhat-actions/oc-installer@v1
+        with:
+          oc_version: 'latest'
+
+      - name: Log in to OpenShift
+        run: |
+          oc login "${{ secrets.TEST_OPENSHIFT_SERVER }}" \
```
ktdreyer, line 76: I'll note one thing here: if we want to deploy to an MP+ cluster behind the firewall, we can use ArgoCD of course, or we can use the _internal GH runner_ as documented in https://github.com/ambient-code/platform/pull/793.

.github/workflows/pr-e2e-openshift.yml
```diff
@@ -0,0 +1,187 @@
...
+          oc login ${{ secrets.TEST_OPENSHIFT_SERVER }} \
+            --token=${{ secrets.TEST_OPENSHIFT_TOKEN }} \
+            --insecure-skip-tls-verify
+
+      - name: Provision namespace
+        run: |
+          bash components/pr-test/provision.sh create \
+            "${{ needs.setup.outputs.instance_id }}"
+
+  install:
```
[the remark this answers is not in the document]
ktdreyer: We must address this

docs/internal/proposals/acp-public-rest-api.md
```diff
@@ -1,214 +0,0 @@
-# Proposal: ACP Public REST API
```
ktdreyer, line 1: Why delete this?

docs/internal/developer/local-development/openshift.md
```diff
@@ -1,382 +1,112 @@
 # OpenShift Cluster Development
 
-This guide covers deploying the Ambient Code Platform on OpenShift clusters for development and testing. Use this when you need to test OpenShift-specific features like Routes, OAuth integration, or service mesh capabilities.
+This guide covers deploying Ambient Code on an OpenShift cluster using the **OpenShift internal image registry**. This is useful when iterating on local builds against a dev cluster without pushing to quay.io.
```
ktdreyer, line 3: Can this change be broken out into a separate PR? I don't understand what this has to do with this change.

components/pr-test/install.sh
```diff
@@ -0,0 +1,202 @@
...
+            doc = re.sub(r'(metadata:)', r'\1\n  labels:', doc, count=1)
+        if 'paas.redhat.com/appcode' not in doc:
+            doc = re.sub(r'(  labels:)', r'\1\n    paas.redhat.com/appcode: AMBC-001', doc, count=1)
+        if 'storageClassName' not in doc:
+            doc = re.sub(r'(spec:)', r'\1\n  storageClassName: aws-ebs', doc, count=1)
+    print('---')
+    print(doc)
+PYEOF
+
+kustomize build . \
```
ktdreyer, line 151:
```suggestion
oc kustomize build . \
```
See https://github.com/ambient-code/platform/issues/768

components/pr-test/README.md
```diff
@@ -0,0 +1,448 @@
+# Specification: Ephemeral PR Test Environments on MPP
```
ktdreyer, line 1: This doc is helpful to understand the design thinking, but some things are already out of date (for example "no branch slug" is wrong). If this stays in this PR, it needs another pass to reconcile with the actual implementation. (See also https://github.com/ambient-code/platform/issues/952)

components/pr-test/install.sh
```diff
@@ -0,0 +1,202 @@
...
+  "quay.io/ambient_code/vteam_frontend:latest=quay.io/ambient_code/vteam_frontend:${IMAGE_TAG}" \
+  "quay.io/ambient_code/vteam_backend:latest=quay.io/ambient_code/vteam_backend:${IMAGE_TAG}" \
+  "quay.io/ambient_code/vteam_operator:latest=quay.io/ambient_code/vteam_operator:${IMAGE_TAG}" \
+  "quay.io/ambient_code/vteam_claude_runner:latest=quay.io/ambient_code/vteam_claude_runner:${IMAGE_TAG}" \
+  "quay.io/ambient_code/vteam_state_sync:latest=quay.io/ambient_code/vteam_state_sync:${IMAGE_TAG}" \
+  "quay.io/ambient_code/vteam_api_server:latest=quay.io/ambient_code/vteam_api_server:${IMAGE_TAG}" \
+  "quay.io/ambient_code/vteam_public_api:latest=quay.io/ambient_code/vteam_public_api:${IMAGE_TAG}"
+
+FILTER_SCRIPT="$TMPDIR/filter.py"
+cat > "$FILTER_SCRIPT" << 'PYEOF'
```
ktdreyer, line 109: This inline Python script is 35 lines of yaml manipulation via regex. That will be very challenging to maintain.

Could this be a standalone Python file in `components/pr-test/filter.py` with proper tests? Or better, could you use `yq` or `oc` instead of regex?

components/pr-test/README.md
ktdreyer: There is significant duplication between the SKILL.md files and this file. For maintainability, I wonder if we could trim these both down and dedup where it makes sense.

docs/internal/developer/local-development/openshift.md
markturansky: by introducing the logic into the skill, it becomes re-usable. it's like a code refactor where we make a change by making a new dual use function and refactoring the old code to use it. this developer guide was my original How To script for claude.

markturansky:
## Review: feat(ci): ephemeral PR test instances on MPP dev cluster

Solid foundational work — the security posture of the workflows (always checking out `main` for scripts, never running PR-supplied code with cluster credentials) is exactly right. The three-script decomposition (`build.sh` / `provision.sh` / `install.sh`) is clean and the skills layer on top naturally.

A few things worth resolving before this lands:

---

### 🔴 Instance ID inconsistency between spec and implementation

`components/pr-test/README.md` § Instance Naming Convention is explicit:
> Instance ID is **PR number only** — no branch slug (avoids namespace name length issues)

But both `pr-e2e-openshift.yml` and `pr-namespace-cleanup.yml` derive:
```
INSTANCE_ID="pr-${PR_NUMBER}-${BRANCH_SLUG}"
```

And the SKILL.md examples use `pr-${PR_NUMBER}` only.

This is a functional bug: if the namespace was provisioned with a branch slug by `pr-e2e-openshift.yml`, but the cleanup workflow computes the same slug independently, they will match — _unless_ the branch name changed (e.g. after a force-push rename) or a race condition occurs. The more pressing issue is that `provision.sh` and the `SKILL.md` both assume `pr-<number>` form, so any manual invocation will compute a different instance ID than CI computed, leaving a namespace stranded.

Pick one form and make it consistent across all four files.

---

### 🟡 `minio-credentials` is still unresolved and will break minio startup

The README documents this as an open item, and `install.sh` correctly omits it from `REQUIRED_SOURCE_SECRETS` (since it doesn't exist in the source namespace). But the production overlay presumably still references this secret, meaning minio pods will fail with `secret "minio-credentials" not found` in every PR namespace. The E2E tests will never reach a healthy state until this is addressed.

This shouldn't block the PR if the intention is to merge infra scaffolding and fix secrets in a follow-up, but it's worth a TODO comment in `install.sh` and an explicit note in the test plan.

---

### 🟡 `destroy()` deletes an ArgoCD Application that likely never exists

```bash
APP_NAME="pr-test-${INSTANCE_ID}"
oc delete application "$APP_NAME" -n "$ARGOCD_NAMESPACE" --ignore-not-found=true
```

This silently no-ops on every run since PR namespaces are deployed via `oc apply` (not ArgoCD). Looks like a leftover from an earlier ArgoCD-managed approach. Safe to remove — the no-op is harmless but misleads anyone reading the teardown path.

---

### 🟡 E2E job always runs `main` test harness, never PR test changes

All four jobs check out `ref: main`. For `provision`, `install`, `teardown` that's intentional and correct. For `e2e` it means new or changed Cypress tests in a PR are never exercised against the ephemeral instance — the old tests on `main` run instead.

If this is intentional (stable harness only), document it. If not, the `e2e` job should check out the PR ref: `ref: ${{ github.event.workflow_run.head_sha }}`.

---

### 🟢 Minor: README secrets table is stale

The table in `README.md` § "Secrets — what must be copied" still marks `postgresql-credentials` and `frontend-oauth-config` as `❌ Not copied` with "add to install.sh" notes — but this PR already adds them to `REQUIRED_SOURCE_SECRETS`. The table should be updated to reflect the current state so it doesn't mislead future readers.

---

Overall the approach is sound and I'm glad the security model was thought through carefully. The namespace ID inconsistency is the only thing that would cause a hard operational failure; the rest can be fixed in follow-ups or addressed before merge depending on how complete you want the first pass to be.

PR #1036, opened 2026-03-25
remove minikube from Makefile

### LINKED ISSUE #992: remove minikube from Makefile
https://github.com/ambient-code/platform/issues/898 dropped most of minikube, but there are still some stray references in the `Makefile`.

This adds complexity to tasks like https://github.com/ambient-code/platform/issues/970, because Claude wants to maintain this unused `make local-sync-version` target.

ktdreyer:
## Summary

- Remove minikube-only targets (`local-url`, `check-minikube`, `_build-and-load`) and minikube references from remaining targets (`local-down`, `local-status`, `local-test-quick`, `makefile-health`, `check-local-context`, `_show-access-info`)
- Add per-component kind reload targets: `kind-reload-backend`, `kind-reload-frontend`, `kind-reload-operator`
- Add deprecation aliases for removed targets (`local-up`, `local-clean`, `local-rebuild`, `local-reload-backend/frontend/operator`) that print a warning and delegate to the kind equivalent
- Follow-up issue #1080 tracks updating docs that still reference the old target names

Fixes #992

## Test plan

- [x] `make help` displays correctly with updated examples
- [x] `make help` still lists `local-up` (with deprecation note)
- [x] `make kind-up` still works as the local dev entrypoint
- [x] `make kind-rebuild` still works for reloading components
- [x] `make kind-reload-backend` / `kind-reload-frontend` / `kind-reload-operator` dry-run succeeds
- [x] `make local-up` prints deprecation warning and delegates to `kind-up`
- [x] `make local-status` correctly detects kind clusters
- [x] `make local-test-quick` correctly detects kind clusters
- [x] `make makefile-health` runs without errors
- [x] No minikube references remain in Makefile (except one comment documenting that the deprecated aliases replace the old minikube-based workflow)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,14 @@
 .PHONY: help setup build-all build-frontend build-backend build-operator build-runner build-state-sync build-public-api build-cli deploy clean check-architecture
-.PHONY: local-up local-down local-clean local-status local-rebuild local-reload-backend local-reload-frontend local-reload-operator local-reload-api-server
+.PHONY: local-down local-status local-reload-api-server local-up local-clean local-rebuild local-reload-backend local-reload-frontend local-reload-operator
 .PHONY: local-dev-token
 .PHONY: local-logs local-logs-backend local-logs-frontend local-logs-operator local-shell local-shell-frontend
-.PHONY: local-test local-test-dev local-test-quick test-all local-url local-troubleshoot local-port-forward local-stop-port-forward
-.PHONY: push-all registry-login setup-hooks remove-hooks lint check-minikube check-kind check-kubectl check-local-context dev-bootstrap kind-rebuild kind-status kind-login
+.PHONY: local-test local-test-dev local-test-quick test-all local-troubleshoot local-port-forward local-stop-port-forward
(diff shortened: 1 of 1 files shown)


ktdreyer:
This is a breaking change. The `test-local-dev-simulation` test failed because `make local-up` no longer works.

I've edited the test.

tests/local-dev-test.sh
```diff
@@ -213,10 +213,9 @@ test_makefile_help() {
     help_output=$(make help 2>&1)
 
     assert_contains "$help_output" "Ambient Code Platform" "Help shows correct branding"
-    assert_contains "$help_output" "local-up" "Help lists local-up command"
+    assert_contains "$help_output" "kind-up" "Help lists kind-up command"
```
ktdreyer: This needs refinement.

Lots of other uses of `make local-up` in the tree at the moment.

ktdreyer: I've resolved this now - we'll keep these around until #1080

bobbravo2 (approved):
Tested this PR locally when working a bug. Kind is solid, thanks 🙏

PR #1058, opened 2026-03-26
feat(backend): support custom SSL root CAs for internal GitLab instances

### LINKED ISSUE #1038: Support custom SSL root CAs for internal GitLab instances
```
```

## Problem

The backend makes outbound HTTPS requests to user-configured GitLab instances during token validation and API calls. All HTTP clients use Go's default TLS configuration, which only trusts the system CA bundle baked into the container image (`ubi9/ubi-minimal`).

This means connections to internal instances that use certificates signed by a corporate CA (e.g. `gitlab.cee.redhat.com`) fail with:

> tls: failed to verify certificate: x509: certificate signed by unknown authority

There is currently no way to provide additional trusted root CAs.

## Affected components

All outbound HTTP clients in the backend that connect to user-configured hosts:

- `components/backend/handlers/integration_validation.go` — `ValidateGitLabToken`, `ValidateGitHubToken`, `ValidateGoogleToken`
- `components/backend/gitlab/client.go` — `NewClient` / `doRequest`
- `components/backend/handlers/gitlab_auth.go` — `validateGitLabConnectivity`

The runner (`components/runners/ambient-runner/`) may also need custom CAs when cloning from internal Git hosts, but that is a separate concern (handled by Git's own CA config).

### Option A: OpenShift trusted CA bundle injection (recommended for OpenShift deployments)

On OpenShift, a ConfigMap labeled with `config.openshift.io/inject-trusted-cabundle: "true"` is automatically populated with the cluster-wide proxy/CA bundle. Mount this into the backend pod and set `SSL_CERT_DIR` or load it into a custom `tls.Config` `RootCAs` pool.

### Option B: Explicit ConfigMap/Secret mount

Allow operators to create a ConfigMap (e.g. `ambient-custom-ca`) containing PEM-encoded CA certificates. Mount it into the backend pod and configure the HTTP clients to include these CAs in their trust pool.

### Implementation notes

- Go respects `SSL_CERT_FILE` and `SSL_CERT_DIR` environment variables for the default TLS config, so the simplest approach may be to mount the CA bundle and set the env var — no code changes needed for existing `http.Client{}` usage.
- If a shared `http.Client` or `http.Transport` is introduced, it should be used consistently across all outbound HTTP calls (GitLab, GitHub, Jira, Google OAuth).
- The LDAP client already has custom TLS handling (`LDAP_SKIP_TLS_VERIFY`) which should *not* serve as a reference, so that we do not spread `InsecureSkipVerify` hacks throughout the codebase.

jeremyeder:
## Summary

- Adds `httputil` package with a shared `http.Transport` that loads additional CA certificates from the `CUSTOM_CA_BUNDLE` environment variable, appending them to the system cert pool
- Updates all outbound HTTP clients connecting to user-configured hosts (`ValidateGitLabToken`, `ValidateGitHubToken`, `ValidateJiraToken`, `ValidateGoogleToken`, `gitlab.NewClient`) to use the shared transport
- Transport mirrors `net/http.DefaultTransport` settings (proxy, HTTP/2, dial timeouts, keep-alive) and falls back gracefully to Go defaults on any CA loading failure

Closes #1038

## Test plan

- [ ] Set `CUSTOM_CA_BUNDLE` to a valid PEM file with a corporate CA and verify connections to an internal GitLab instance succeed
- [ ] Verify connections to public hosts (gitlab.com, github.com, googleapis.com) still work without `CUSTOM_CA_BUNDLE` set
- [ ] Set `CUSTOM_CA_BUNDLE` to a nonexistent path and verify the backend starts with a warning log and falls back to system CAs
- [ ] Set `CUSTOM_CA_BUNDLE` to a file with invalid PEM content and verify the backend starts with a warning and falls back to system CAs
- [ ] Verify `HTTP_PROXY`/`HTTPS_PROXY` environment variables are still respected

### DIFF
--- a/components/backend/gitlab/client.go
+++ b/components/backend/gitlab/client.go
@@ -12,6 +12,7 @@ import (
 	"strconv"
 	"time"
 
+	"ambient-code-backend/httputil"
 	"ambient-code-backend/types"
 	"github.com/google/uuid"
 )
@@ -31,11 +32,9 @@ type Client struct {
 // NewClient creates a new GitLab API client with 15-second timeout
 func NewClient(baseURL, token string) *Client {
 	return &Client{
-		httpClient: &http.Client{
-			Timeout: 15 * time.Second,
-		},
-		baseURL: baseURL,
-		token:   token,
+		httpClient: httputil.NewClient(15 * time.Second),
+		baseURL:    baseURL,
+		token:      token,
 	}
 }
 
--- a/components/backend/handlers/integration_validation.go
+++ b/components/backend/handlers/integration_validation.go
@@ -6,6 +6,8 @@ import (
 	"net/http"
 	"time"
 
+	"ambient-code-backend/httputil"
+
 	"github.com/gin-gonic/gin"
 )
 
@@ -15,7 +17,7 @@ func ValidateGitHubToken(ctx context.Context, token string) (bool, error) {
 		return false, fmt.Errorf("token is empty")
 	}
 
-	client := &http.Client{Timeout: 10 * time.Second}
+	client := httputil.NewClient(10 * time.Second)
 	req, err := http.NewRequestWithContext(ctx, "GET", "https://api.github.com/user", nil)
 	if err != nil {
 		return false, fmt.Errorf("failed to create request")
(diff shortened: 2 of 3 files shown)


ktdreyer:
What do you think about the first option (`config.openshift.io/inject-trusted-cabundle: "true"`)? That seems more elegant, if it works.

ktdreyer:
I tried `config.openshift.io/inject-trusted-cabundle: "true"` recently for ArgoCD. They already pre-populate it with the RH CA in RH IT's MP+ environment. So that is definitely the way to go.

ktdreyer:
I think we can close this. I'm moving forward with the other solutions linked in the parent issue 1038

Issue #1024, opened 2026-03-25
Bug: Firefox start session box doesn't expand


jwm4:
## 🐛 Bug Description

In firefox, when I'm starting a new session and entering multiple lines in the prompt box it doesn't expand and the text overlaps the control elements.

## 🖼️ Screenshots

Here you can see the problem in Firefox:

<img width="2276" height="838" alt="Image" src="https://github.com/user-attachments/assets/82d9af19-b1d2-4d77-a764-1c3b3dff3765" />

Here you can see that the problem does NOT exist in Chrome:

<img width="1570" height="740" alt="Image" src="https://github.com/user-attachments/assets/efb3c34d-f242-4d6b-8ba3-4c026af300be" />

In Chrome, the box expands as needed.

## ✅ Acceptance Criteria

- [ ] Bug is reproduced and root cause identified
- [ ] Fix is implemented and tested
- [ ] Regression tests added to prevent future occurrences
- [ ] Documentation updated if needed
- [ ] Fix is verified in staging environment

## 🏷️ Labels

- **Priority:** [low/medium/high/critical]
- **Complexity:** [trivial/easy/medium/hard]
- **Component:** [frontend/backend/operator/tools/docs]

dgutride:
Fixed :)

PR #1061, opened 2026-03-26
fix: Firefox session layout was clipping content

dgutride:
Fixes firefox layout issues - and an issue I Noticed when filling the entire screen with content where it slid under the top header:

Screenshot 2026-03-26 at 2 36 25 PM


dgutride:
Fixes https://github.com/ambient-code/platform/issues/1024

PR #1105, opened 2026-03-30
fix: resolve Docker image prefix mismatch for kind-local deploys

jeremyeder:
## Summary

- `KIND_IMAGE_PREFIX` was only set to `localhost/` for podman, but empty for Docker
- The `kind-local` kustomize overlay always references `localhost/vteam_*` images
- With Docker, `make kind-up LOCAL_IMAGES=true` loaded unprefixed images, causing `ImagePullBackOff` on every deploy

**Fix:** Always set `KIND_IMAGE_PREFIX` to `localhost/` and tag Docker-built images before loading into kind.

## Test plan

- [x] Verified fix resolves `ImagePullBackOff` when using `CONTAINER_ENGINE=docker`
- [ ] `make kind-up LOCAL_IMAGES=true CONTAINER_ENGINE=docker` completes without image pull errors
- [ ] `make kind-up LOCAL_IMAGES=true CONTAINER_ENGINE=podman` still works (no behavioral change)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,9 +66,9 @@ STATE_SYNC_IMAGE ?= vteam_state_sync:$(IMAGE_TAG)
 PUBLIC_API_IMAGE ?= vteam_public_api:$(IMAGE_TAG)
 API_SERVER_IMAGE ?= vteam_api_server:$(IMAGE_TAG)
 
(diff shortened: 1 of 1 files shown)


jeremyeder:
I hit this every time I use kind with locally built images.

PR #1110, opened 2026-03-31
docs(spec): Credential Kind — design spec for RHOAIENG-55817

markturansky:
## Summary

- Introduces `Credential` as a first-class Ambient Kind in the platform data model spec
- Adds `docs/internal/design/credentials-session.md` — full design spec with ERD, ownership model, API/CLI reference, usage examples, and open questions
- Updates `docs/internal/design/ambient-model.spec.md` to reflect the desired state: new Credential entity, `RoleBinding.scope` extended with `credential`, new roles (`credential:owner`, `credential:reader`), Credentials API and CLI sections

## What this is

This is a **spec-only PR** — no code changes. The goal is design review before implementation begins. The reconciler will use this spec as the desired state to surface implementation gaps.

Key design decisions captured:
- `Credential` is platform-scoped (not project/agent-scoped) to support shared Robot Accounts
- Ownership via `RoleBinding(scope=credential, role=credential:owner)` — consistent with Agent ownership pattern
- Token is write-only; never returned via standard REST API
- Scope hierarchy (agent → project → global) for credential resolution at session ignition
- Runner token endpoint shape is marked TBD (open question in the design doc)

## Test plan
- [ ] Design review — read `docs/internal/design/credentials-session.md`
- [ ] Verify ERD changes in `ambient-model.spec.md` are consistent with the design doc
- [ ] Answer open questions in `credentials-session.md` before implementation begins

Closes RHOAIENG-55817 (design phase)

### DIFF
--- a/components/ambient-cli/cmd/acpctl/credential/cmd.go
+++ b/components/ambient-cli/cmd/acpctl/credential/cmd.go
@@ -1,0 +1,406 @@
+package credential
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/ambient-code/platform/components/ambient-cli/pkg/config"
+	"github.com/ambient-code/platform/components/ambient-cli/pkg/connection"
+	"github.com/ambient-code/platform/components/ambient-cli/pkg/output"
+	sdktypes "github.com/ambient-code/platform/components/ambient-sdk/go-sdk/types"
+	"github.com/spf13/cobra"
+)
+
+var Cmd = &cobra.Command{
+	Use:   "credential",
+	Short: "Manage credentials",
+	Long: `Manage credentials for external service integrations.
+
+Subcommands:
+  list        List credentials
+  get         Get a specific credential
+  create      Create a credential
+  update      Update a credential's fields
+  delete      Delete a credential
+  token       Retrieve the token for a credential`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return cmd.Help()
+	},
+}
+
+var listArgs struct {
+	outputFormat string
+	limit        int
+	provider     string
+}
+
+var listCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List credentials",
+	Example: `  acpctl credential list
+  acpctl credential list --provider github -o json`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		client, err := connection.NewClientFromConfig()
+		if err != nil {
+			return err
+		}
+
+		cfg, err := config.Load()
+		if err != nil {
+			return err
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), cfg.GetRequestTimeout())
+		defer cancel()
+
+		format, err := output.ParseFormat(listArgs.outputFormat)
+		if err != nil {
+			return err
+		}
+		printer := output.NewPrinter(format, cmd.OutOrStdout())
+
+		opts := sdktypes.NewListOptions().Size(listArgs.limit).Build()
+		if listArgs.provider != "" {
+			opts.Search = fmt.Sprintf("provider='%s'", listArgs.provider)
+		}
+		list, err := client.Credentials().List(ctx, opts)
+		if err != nil {
+			return fmt.Errorf("list credentials: %w", err)
+		}
+
+		if printer.Format() == output.FormatJSON {
+			return printer.PrintJSON(list)
+		}
+
+		return printCredentialTable(printer, list.Items)
+	},
+}
+
+var getArgs struct {
+	outputFormat string
+}
+
+var getCmd = &cobra.Command{
+	Use:   "get <id>",
+	Short: "Get a specific credential",
+	Args:  cobra.ExactArgs(1),
+	Example: `  acpctl credential get <id>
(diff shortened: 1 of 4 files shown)


docs/internal/design/ambient-model.spec.md
```diff
@@ -529,6 +566,42 @@ GET    /api/ambient/v1/sessions/{id}/events                  SSE AG-UI event str
...
+GET    /api/ambient/v1/credentials                           list credentials visible to the caller
+GET    /api/ambient/v1/credentials?provider={provider}       filter by provider
+POST   /api/ambient/v1/credentials                           create a credential
+GET    /api/ambient/v1/credentials/{id}                      read credential (metadata only; token never returned)
+PATCH  /api/ambient/v1/credentials/{id}                      update credential
+DELETE /api/ambient/v1/credentials/{id}                      soft delete
+GET    /api/ambient/v1/credentials/{id}/role_bindings        RBAC bindings on this credential
+```
+
+`token` is accepted on `POST` and `PATCH` but **never returned** in any response. It is stored encrypted in the database. The database row is the authoritative store; a future Vault integration can be adopted by pointing the row at a Vault path without changing the API surface.
```
markturansky: adding bit on how the runner securely obtains the token

docs/internal/design/ambient-model.spec.md
```diff
@@ -529,6 +566,42 @@ GET    /api/ambient/v1/sessions/{id}/events                  SSE AG-UI event str
...
+
+#### Provider Enum
+
+| Provider | Service | Token type | `url` | `email` |
+|----------|---------|------------|-------|---------|
+| `github` | GitHub.com or GitHub Enterprise | Personal Access Token | optional; required for GHE | — |
+| `gitlab` | GitLab.com or self-hosted | Personal Access Token | optional; required for self-hosted | — |
+| `jira` | Jira Cloud (Atlassian) | API Token | required (Atlassian instance URL) | required (used in Basic auth) |
+| `google` | Google Cloud / Workspace | Service Account JSON serialized to string | — | — |
```
markturansky, line 594: adding `other` as a catch-up generic provider

docs/internal/design/ambient-model.spec.md
```diff
@@ -142,13 +148,30 @@ erDiagram
...
         time   deleted_at
     }
 
+    %% ── Credential (platform-scoped PAT/token store) ─────────────────────────
+
+    Credential {
+        string ID PK "KSUID"
+        string name "human-readable"
+        string description
+        string provider "github | gitlab | jira | google | other"
```
jwm4: The "other" case seems kind of tricky to me.  There are a lot of "other" credentials for other things that we might want to connect to.  Some things require tokens.  Some require API keys.  Some require username and passwords.  Occasionally, you get something exotic like RH SSO requiring a username plus a pin plus an OTP.  Maybe it's OK to scope this for just things that require tokens.  Or maybe it's OK to scope it for things that require tokens and also things that require API keys and if you have something that requires an API key you just to get into the token field even though technically it's not a token -- it seems like a pretty common pattern with the APIs.

It might be helpful to start with an MVP set of examples of "other" providers, we want to support and then see what authentication they require and check whether the spec can handle what they need.

docs/internal/design/ambient-model.spec.md
```diff
@@ -529,6 +567,47 @@ GET    /api/ambient/v1/sessions/{id}/events                  SSE AG-UI event str
...
+#### Token Response Shape (Runner)
+
+When a runner fetches a credential, the response payload shape is consistent across providers:
+
+```json
+{ "provider": "gitlab", "token": "glpat-...",       "url": "https://gitlab.myco.com" }
+{ "provider": "github", "token": "github_pat_...",  "url": "https://github.com" }
+{ "provider": "jira",   "token": "ATATT3x...",      "url": "https://myco.atlassian.net", "email": "bot@myco.com" }
+{ "provider": "google", "token": "{\"type\":\"service_account\", ...}" }
+{ "provider": "other",  "token": "...", "url": "https://my-service.example.com" }
```
jwm4: I'm not totally clear what the runner does with this information about this "other" example.  It has a URL from this object, but the object isn't telling me whether this is a REST server or a gRPC server or an MCP server or something else.  It isn't telling me what the end points are or what the scheme is are for those end points.  So I'm not clear on what the runner is really going to do with that.

docs/internal/design/ambient-model.spec.md
markturansky: you're right.  the google token is a complex json structure while the others are just a token.

Is it good enough to mount the secret if the user asks for it and they have their own agent do stuff with it?  they define their own curl commands and so on in their prompt.

docs/internal/design/ambient-model.spec.md
markturansky: similar to the other comment, I'm presuming it's enough to mount the secret the user asked for and let them have their agent do something with it.

markturansky:
@jwm4 per our convo in slack, removing 'other'. that use case will be covered with BYOMCP.

markturansky:
'other' removed. tests are green. validated locally. merging.

PR #1117, opened 2026-03-31
spec: skills and workflows — discovery, installation, and usage

Gkrumbach07:
## Summary

Defines the architecture for how skills, workflows, and agents work in ACP.

**Core insight**: A workflow is just a prompt + a list of skill source references. Skills are the atomic reusable unit. The manifest format is the contract — ACP and local tooling are just consumers.

## Key Decisions

- **Workflows don't contain skills** — they reference them by Git URL + path
- **Skills are always fetched from canonical sources** at load time, not embedded copies
- **Workspace library != auto-inject** — installing to workspace makes items available, not active. Users select what loads per session.
- **Same manifest works everywhere** — ACP runner, local `--add-dir`, a `/load-workflow` meta-skill, or a Claude Code plugin
- **Layered loading** — workflow dependencies + user-selected extras + live imports, all fed into `add_dirs`

## Sections

1. **Core Concepts** — Skill, Workflow, Agent, Manifest Format
2. **Discovery** — Marketplace, scanning, catalog format
3. **Installation & Configuration** — Workspace level, session level, workflow builder, selection model
4. **Usage in Sessions** — Loading order, runtime management, metadata
5. **Local Usage** — Manual, load-workflow skill, Claude Code plugin
6. **Manifest Format** — Proposed YAML schema
7. **Open Questions** — Format, versioning, plugin compat, RHAI alignment, security

## For Review

Looking for feedback on:
- The manifest format (extend ambient.json vs new workflow.yaml?)
- The selection model (workspace library vs auto-inject)
- The Agent concept (persona + workflows + skills)
- RHAI alignment (RHAIRFE-1370)

### DIFF
--- a/specs/skills-and-workflows.md
+++ b/specs/skills-and-workflows.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-A workflow is a prompt plus a list of skill sources. Skills are the atomic reusable unit. ACP automates the cloning and wiring; locally, the same manifest format works with a simple skill or script. The manifest format is the contract — ACP and local tooling are just consumers.
+A workflow is a `CLAUDE.md` prompt plus a list of sources in `ambient.json`. Skills are the atomic reusable unit. ACP automates the cloning and wiring. Locally, a `/load-workflow` skill or manual `--add-dir` does the same thing.
 
 ---
 
@@ -20,38 +20,52 @@ Commands (`.claude/commands/{name}.md`) and agents (`.claude/agents/{name}.md`) 
 
 A workflow is two things:
 
-1. **A prompt** — the directive, methodology, persona instructions. What today lives in `CLAUDE.md` + `.ambient/ambient.json` (system prompt, startup prompt, phase descriptions).
-2. **A list of skill sources** — references to skills, commands, and agents from various Git repos or registries. Not embedded copies — references resolved at load time.
+1. **A prompt** — the directive and methodology, written as `CLAUDE.md` in the workflow directory. This is the only prompt mechanism — no separate `systemPrompt` or `startupPrompt` fields.
+2. **A list of sources** — references in `ambient.json` to skills, commands, agents, and plugins from various Git repos. Not embedded copies — references resolved at load time.
 
 A workflow does not contain skills. It references them. The bug-fix workflow becomes:
 
-```yaml
-name: Bug Fix
-description: Systematic bug resolution with phased approach
-prompt: |
-  You are a systematic bug fixer. Follow these phases:
-  1. Use /assess to understand the issue
-  2. Use /reproduce to create a failing test
-  3. Use /diagnose to find the root cause
-  4. Use /fix to implement the minimal fix
-  5. Use /test to verify the fix
-  6. Use /review to self-review before PR
-sources:
-  - url: https://github.com/ambient-code/skills.git
-    path: bugfix/assess
-  - url: https://github.com/ambient-code/skills.git
-    path: bugfix/reproduce
-  - url: https://github.com/ambient-code/skills.git
-    path: bugfix/diagnose
-  - url: https://github.com/ambient-code/skills.git
-    path: bugfix/fix
-  - url: https://github.com/ambient-code/skills.git
-    path: bugfix/test
-  - url: https://github.com/ambient-code/skills.git
-    path: bugfix/review
-  - url: https://github.com/opendatahub-io/ai-helpers.git
-    path: helpers/skills/jira-activity
+**`CLAUDE.md`**:
+```markdown
+You are a systematic bug fixer. Follow these phases:
+1. Use /assess to understand the issue
+2. Use /reproduce to create a failing test
+3. Use /diagnose to find the root cause
+4. Use /fix to implement the minimal fix
+5. Use /test to verify the fix
+6. Use /review to self-review before PR
 ```
+
+**`ambient.json`**:
+```json
+{
+  "name": "Bug Fix",
+  "description": "Systematic bug resolution with phased approach",
+  "sources": [
+    {"url": "https://github.com/ambient-code/skills.git", "branch": "main", "path": "bugfix/assess"},
+    {"url": "https://github.com/ambient-code/skills.git", "path": "bugfix/reproduce"},
+    {"url": "https://github.com/ambient-code/skills.git", "path": "bugfix/diagnose"},
+    {"url": "https://github.com/ambient-code/skills.git", "path": "bugfix/fix"},
+    {"url": "https://github.com/ambient-code/skills.git", "path": "bugfix/test"},
+    {"url": "https://github.com/ambient-code/skills.git", "path": "bugfix/review"},
+    {"url": "https://github.com/opendatahub-io/ai-helpers.git", "path": "helpers/skills/jira-activity"},
+    "https://github.com/my-org/shared-skills/tree/main/code-review"
+  ],
+  "rubric": {
+    "activationPrompt": "After completing the fix, evaluate your work",
+    "criteria": [
+      {"name": "Root cause identified", "weight": 0.3},
+      {"name": "Tests added", "weight": 0.3},
+      {"name": "Minimal change", "weight": 0.2},
+      {"name": "No regressions", "weight": 0.2}
+    ]
+  }
+}
+```
+
+Sources support two formats:
+- **Structured object**: `{"url": "...", "branch": "...", "path": "..."}` — works with any Git host, branch is explicit
+- **Single URL string**: `"https://github.com/org/repo/tree/main/path"` — auto-parsed, convenient for sharing
 
 Skills are the reusable atoms. Workflows are recipes. The same skill can appear in multiple workflows.
 
@@ -60,24 +74,12 @@ Skills are the reusable atoms. Workflows are recipes. The same skill can appear 
 A persona — prose defining what an agent is responsible for. "Backend Agent", "Security Agent", "PM Agent". An Agent uses workflows and standalone skills to accomplish its goals. An Agent is a session template with a personality:
 
 ```
-Agent = Persona (prompt/directive) + Workflows (skill bundles) + Standalone skills
+Agent = Persona (CLAUDE.md) + Workflows (skill bundles) + Standalone skills
 ```
 
 A "Bug Fix Agent" = bug-fix persona + bug-fix workflow skills + any additional skills. Same skills reusable by different Agents with different motivations.
 
-### Manifest Format
-
-The workflow manifest (`workflow.yaml` or `ambient.json`) is the core deliverable. It defines prompt + skill sources. Multiple consumers read the same format:
-
-- **ACP runner**: clones sources, adds to `add_dirs`, injects prompt
-- **Local script/skill**: clones sources to temp dirs, passes as `--add-dir`
-- **Claude Code plugin**: `SessionStart` hook reads manifest, sets up environment
-
-The format must support:
-- Prompt text (inline or file reference)
-- Skill sources (Git URL + branch + path)
-- Optional: system prompt, startup prompt, rubric
-- Optional: environment variables, MCP server configs
+Multi-agent orchestration (research agent → writer agent → editor agent pipelines) is a separate design problem out of scope for this spec.
 
 ---
 
@@ -87,19 +89,23 @@ The format must support:
 
 A way to browse and find skills, workflows, and plugins from curated sources.
 
+### Source Types
+
+The scanner must support three types of sources:
+
+1. **Claude Code plugins** — directories with `.claude-plugin/plugin.json` containing `skills/`, `commands/`, `agents/`, `hooks/`, `.mcp.json`. This is the primary format to follow and expect. Skills are namespaced as `plugin-name:skill-name`.
+
+2. **Claude Code marketplace catalogs** — `marketplace.json` files listing plugins with their sources. Users could add the same marketplace from local Claude Code via `/plugin marketplace add`.
+
+3. **Standalone repos with `.claude/`** — any Git repo containing `.claude/skills/`, `.claude/commands/`, `.claude/agents/`. Also supports root-level `skills/`, `commands/`, `agents/` (registry layout like ai-helpers).
+
 ### How
 
-A cluster-level ConfigMap (`marketplace-sources`) holds available registry sources. Each source is one of:
+A cluster-level ConfigMap (`marketplace-sources`) holds available registry sources. The Marketplace page in the ACP UI shows:
 
-- A Git repo with a `data.json` catalog (like ai-helpers at `https://opendatahub-io.github.io/ai-helpers/data.json`)
-- A Claude Code `marketplace.json` (native plugin marketplace format)
-- A direct Git repo with skills/commands/agents in `.claude/` or at root level
-
-The Marketplace page in the ACP UI shows:
-
-- Browsable catalogs from each source with search and type filters (skill / command / agent / workflow)
+- Browsable catalogs from each source with search and type filters
 - Compact card tiles with name, description, type badge
-- Detail panel on click with full description, source repo link, allowed tools
+- Detail panel on click with full description, source repo link
 - "Import Custom" to scan any Git URL and discover items
 - Direct one-click install to workspace
 
@@ -109,16 +115,17 @@ When given a Git URL (from marketplace or custom), the backend:
 
 1. Shallow clones the repo
 2. Applies optional path filter (subdirectory)
-3. Scans for items in both patterns:
+3. Checks for `.claude-plugin/plugin.json` (Claude Code plugin format)
+4. Scans for items in both patterns:
    - `.claude/skills/*/SKILL.md`, `.claude/commands/*.md`, `.claude/agents/*.md`
-   - `skills/*/SKILL.md`, `commands/*.md`, `agents/*.md` (registry layout like ai-helpers)
-4. Checks for `.ambient/ambient.json` (indicates this is a workflow)
-5. Checks for `CLAUDE.md` (indicates project instructions)
-6. Returns discovered items with frontmatter metadata
+   - `skills/*/SKILL.md`, `commands/*.md`, `agents/*.md` (registry layout)
+5. Checks for `.ambient/ambient.json` (indicates this is a workflow)
+6. Checks for `CLAUDE.md` (indicates project instructions)
+7. Returns discovered items with frontmatter metadata
 
-### Alignment with Claude Code
+### Format Alignment
 
-Adopting Claude Code's `marketplace.json` format where possible makes ACP skills portable. Users could add the same marketplace from local Claude Code via `/plugin marketplace add`. The catalog format should normalize to the same shape regardless of source type.
+We follow Claude Code's plugin and skill formats as the standard. The [Agent Skills](https://agentskills.io) open standard that Claude Code implements is the closest cross-tool specification. Our catalog format normalizes to the same shape regardless of source type.
 
 ---
 
@@ -128,7 +135,7 @@ Adopting Claude Code's `marketplace.json` format where possible makes ACP skills
 
 Items installed at the workspace level are stored in the ProjectSettings CR (`spec.installedItems`). These represent the workspace's **library** — what's available, not what's auto-loaded into every session.
 
-When creating a session, users select which installed items to include. The workflow they choose may also pull in its own skill dependencies from the manifest.
+When creating a session, users select which installed items to include. The workflow they choose pulls in its own skill dependencies from the `sources` array in `ambient.json`.
 
 ### Session Level
 
@@ -146,21 +153,22 @@ A UI for composing workflows from standalone skills:
 
 - Select skills from the workspace library or browse marketplace
 - Each skill is a reference (source URL + path), not a copy
-- Define the workflow prompt (methodology, phases, instructions)
-- Optionally configure: system prompt, startup prompt, rubric
-- Save as a workflow manifest that can be:
+- Write the workflow prompt as `CLAUDE.md`
+- Configure metadata in `ambient.json` (name, description, rubric)
+- The `sources` array is built from selected skills
+- Save as a workflow that can be:
   - Stored in the workspace
   - Exported as a Git repo
-  - Exported as a Claude Code plugin (`plugin.json`)
+  - Exported as a Claude Code plugin
 
-The key constraint: skills are never copied into the workflow. The manifest holds references. At load time, the runner resolves dependencies and clones each source.
+The key constraint: skills are never copied into the workflow. The `ambient.json` holds source references. At load time, the runner resolves dependencies and clones each source.
 
 ### How Selection Works
 
 The workspace library is not auto-injected. Selection happens at session creation:
 
 1. User picks a workflow (or "General chat" for none)
-2. The workflow manifest declares its skill dependencies — those are auto-loaded
+2. The workflow's `ambient.json` `sources` array declares dependencies — those are auto-loaded
 3. User can optionally add standalone skills from the workspace library
 4. The session CRD stores the workflow reference + any additional skill sources
 
@@ -178,13 +186,17 @@ This means:
 
 When a session starts, sources are loaded in layers:
 
-1. **Workflow sources** — skills referenced in the workflow manifest, cloned and added to `add_dirs`
+1. **Workflow sources** — skills from the workflow's `ambient.json` `sources` array, cloned and added to `add_dirs`
 2. **Additional standalone sources** — extra skills the user selected at session creation
 3. **Live additions** — skills imported during the session via the context panel
 
 All layers produce directories with `.claude/skills/`, `.claude/commands/`, `.claude/agents/` structure. Each directory is passed to the Claude Agent SDK as an `--add-dir`. Claude Code handles discovery from there.
 
-The `CLAUDE_CODE_ADDITIONAL_DIRECTORIES_CLAUDE_MD=1` env var is set so that `CLAUDE.md` files from add-dirs are also loaded (workflow instructions, skill documentation).
+The `CLAUDE_CODE_ADDITIONAL_DIRECTORIES_CLAUDE_MD=1` env var is set so that `CLAUDE.md` files from add-dirs are also loaded.
+
+### Authentication for Sources
+
+Private repos and authenticated services (MCP servers) use the existing workspace credential system. If the workspace has GitHub/GitLab integrations configured, private source repos are cloned using those credentials via the git credential helper. MCP sources that require auth and TLS are handled through workspace integration configuration. No new auth fields in the manifest.
 
 ### Runtime Management
 
@@ -210,9 +222,7 @@ The frontend uses this to populate the Skills toolbar button and `/` autocomplet
 
 ## Local Usage (outside ACP)
 
-The same workflow manifest works locally. Two approaches:
-
-### 1. Manual (no tooling)
+### 1. Manual
 
 ```bash
 git clone --depth 1 https://github.com/ambient-code/skills.git /tmp/skills
@@ -225,7 +235,7 @@ claude \
 
 ### 2. Load-workflow skill
 
-A meta-skill that reads a workflow manifest and sets up the environment:
+A meta-skill that reads a workflow's `ambient.json`, clones each source, and passes them as `--add-dir`:
 
 ```
 ~/.claude/skills/load-workflow/SKILL.md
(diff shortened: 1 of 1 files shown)


specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+
+Skills have live change detection in `--add-dir` directories — place a new skill file and Claude discovers it immediately without restarting. Skills are invoked via `/skill-name` or auto-triggered by Claude based on the description in frontmatter.
+
+Commands (`.claude/commands/{name}.md`) and agents (`.claude/agents/{name}.md`) follow the same discovery pattern and are treated as peers to skills throughout this spec. "Skill" is used as shorthand for all three unless distinction matters.
+
+### Workflow
+
+A workflow is two things:
+
+1. **A prompt** — the directive, methodology, persona instructions. What today lives in `CLAUDE.md` + `.ambient/ambient.json` (system prompt, startup prompt, phase descriptions).
```
Gkrumbach07:
```suggestion
1. **A prompt** — the directive, methodology, persona instructions. What today lives in `CLAUDE.md` + `.ambient/ambient.json` (system prompt, startup prompt, rubric).
```

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+    path: bugfix/diagnose
+  - url: https://github.com/ambient-code/skills.git
+    path: bugfix/fix
+  - url: https://github.com/ambient-code/skills.git
+    path: bugfix/test
+  - url: https://github.com/ambient-code/skills.git
+    path: bugfix/review
+  - url: https://github.com/opendatahub-io/ai-helpers.git
+    path: helpers/skills/jira-activity
+```
```
Gkrumbach07: this sources should just be part of ambient.json also we need branch

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+
+### Agent (future)
+
+A persona — prose defining what an agent is responsible for. "Backend Agent", "Security Agent", "PM Agent". An Agent uses workflows and standalone skills to accomplish its goals. An Agent is a session template with a personality:
+
+```
+Agent = Persona (prompt/directive) + Workflows (skill bundles) + Standalone skills
+```
+
+A "Bug Fix Agent" = bug-fix persona + bug-fix workflow skills + any additional skills. Same skills reusable by different Agents with different motivations.
```
Gkrumbach07, line 80: remove this

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+
+### Manifest Format
+
+The workflow manifest (`workflow.yaml` or `ambient.json`) is the core deliverable. It defines prompt + skill sources. Multiple consumers read the same format:
```
Gkrumbach07: no workflow.yaml

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+
+- **ACP runner**: clones sources, adds to `add_dirs`, injects prompt
```
Gkrumbach07: we already inject prompt

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+- **Local script/skill**: clones sources to temp dirs, passes as `--add-dir`
+- **Claude Code plugin**: `SessionStart` hook reads manifest, sets up environment
```
Gkrumbach07: we donmt need to use a hook we can just use prestart stuff in hydrate or in the pre start script in the container

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+```
+
+A "Bug Fix Agent" = bug-fix persona + bug-fix workflow skills + any additional skills. Same skills reusable by different Agents with different motivations.
+
+### Manifest Format
+
+The workflow manifest (`workflow.yaml` or `ambient.json`) is the core deliverable. It defines prompt + skill sources. Multiple consumers read the same format:
+
+- **ACP runner**: clones sources, adds to `add_dirs`, injects prompt
+- **Local script/skill**: clones sources to temp dirs, passes as `--add-dir`
```
Gkrumbach07:
```suggestion
- **Local skill**: clones sources to temp dirs, passes as `--add-dir`
```

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+- **Claude Code plugin**: `SessionStart` hook reads manifest, sets up environment
+
+The format must support:
+- Prompt text (inline or file reference)
+- Skill sources (Git URL + branch + path)
+- Optional: system prompt, startup prompt, rubric
+- Optional: environment variables, MCP server configs
```
Gkrumbach07: we are just adding sources

specs/skills-and-workflows.md
jwm4: I like the idea of collapsing what today lives in `CLAUDE.md` + `ambient.json` (system prompt, startup prompt, phase descriptions) into a single prompt. However, I'd much rather have that single prompt live in a markdown file in most cases than in a YAML string.  So I guess I'd prefer something more like what's done with sources where you don't embedded the prompt into this YAML -- instead the YAML just has a pointer to a markdown file.  On the other hand, I can certainly see that they are probably cases out there where you want to use a workflow that somebody already had built somewhere else and they haven't provided a CLAUDE.md or ambient.json or both for it, and then it's much more convenient to just stick the prompt into the manifest.  Maybe we wind up giving both options?  e.g., the schema could have both prompt and prompt_url fields and you can set one or the other, but not both.

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+  You are a systematic bug fixer. Follow these phases:
+  1. Use /assess to understand the issue
+  2. Use /reproduce to create a failing test
+  3. Use /diagnose to find the root cause
+  4. Use /fix to implement the minimal fix
+  5. Use /test to verify the fix
+  6. Use /review to self-review before PR
+sources:
+  - url: https://github.com/ambient-code/skills.git
+    path: bugfix/assess
```
jwm4: URL + path seems clunky.  I'd prefer a single URL that goes directly to the source, e.g., `https://github.com/ambient-code/workflows/blob/main/workflows/bugfix/.claude/skills/assess/`

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+description: Systematic bug resolution with phased approach
+prompt: |
+  You are a systematic bug fixer. Follow these phases:
+  1. Use /assess to understand the issue
+  2. Use /reproduce to create a failing test
+  3. Use /diagnose to find the root cause
+  4. Use /fix to implement the minimal fix
+  5. Use /test to verify the fix
+  6. Use /review to self-review before PR
+sources:
```
jwm4: Should this be called `skills` instead of `sources`?  I feel like maybe we want other sources besides skills but if so, maybe they should have their own top level grouping.  Or maybe there should be a source type on each source indicate whether it's a skill or something else.  Source types could include skills, rules, commands, hooks, etc.

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+    path: bugfix/review
+  - url: https://github.com/opendatahub-io/ai-helpers.git
+    path: helpers/skills/jira-activity
+```
+
+Skills are the reusable atoms. Workflows are recipes. The same skill can appear in multiple workflows.
+
+### Agent (future)
+
+A persona — prose defining what an agent is responsible for. "Backend Agent", "Security Agent", "PM Agent". An Agent uses workflows and standalone skills to accomplish its goals. An Agent is a session template with a personality:
```
jwm4, line 74: Agents could also be sources presumably, right?

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+
+```
+Agent = Persona (prompt/directive) + Workflows (skill bundles) + Standalone skills
```
jwm4: This formula does make sense, but I think there's also a lot of design around workflows that include multiple agents, e.g., a workflow for creating a report that includes a research agent, followed by a writer agent, followed by an editor agent, followed by a distributor agent.  I think it's going to be hard for us to have a concept of an agent here that fits well with all the different things that people want to do with multiple agents.

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+3. Scans for items in both patterns:
+   - `.claude/skills/*/SKILL.md`, `.claude/commands/*.md`, `.claude/agents/*.md`
+   - `skills/*/SKILL.md`, `commands/*.md`, `agents/*.md` (registry layout like ai-helpers)
+4. Checks for `.ambient/ambient.json` (indicates this is a workflow)
+5. Checks for `CLAUDE.md` (indicates project instructions)
+6. Returns discovered items with frontmatter metadata
+
+### Alignment with Claude Code
+
+Adopting Claude Code's `marketplace.json` format where possible makes ACP skills portable. Users could add the same marketplace from local Claude Code via `/plugin marketplace add`. The catalog format should normalize to the same shape regardless of source type.
```
jwm4: That sounds reasonable to me.  I'd really prefer to align with a non-vendor-specific equivalent, but I don't think there is one.

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+
+### What
+
+A way to browse and find skills, workflows, and plugins from curated sources.
+
+### How
+
+A cluster-level ConfigMap (`marketplace-sources`) holds available registry sources. Each source is one of:
+
+- A Git repo with a `data.json` catalog (like ai-helpers at `https://opendatahub-io.github.io/ai-helpers/data.json`)
```
Gkrumbach07: so ai helpers is actually not correct

look here: https://github.com/anthropics/claude-plugins-official/blob/main/plugins/claude-code-setup/README.md, https://code.claude.com/docs/en/skills#share-skills, https://code.claude.com/docs/en/plugins

basically we need to look for certain ways skills are stored and used 

basically plugins are a dir that u can load in. they have a plugin.json and a marketplace.json. skills live within a .claude/skills simielr for xcommands

we need to folloiwe and expect these fortmats

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+  - url: https://github.com/ambient-code/skills.git
+    path: bugfix/reproduce
+  - url: https://github.com/opendatahub-io/ai-helpers.git
+    path: helpers/skills/jira-activity
+  - url: https://github.com/my-org/internal-skills.git
+    branch: v2
+    path: code-review
+
+# Optional: quality evaluation criteria
+rubric:
```
jwm4: I guess I'm increasingly inclined to drop this.  Having it in our schema encourages people to use it and then they wind up making things that are less portable because they depend on it.

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+
+A cluster-level ConfigMap (`marketplace-sources`) holds available registry sources. Each source is one of:
+
+- A Git repo with a `data.json` catalog (like ai-helpers at `https://opendatahub-io.github.io/ai-helpers/data.json`)
+- A Claude Code `marketplace.json` (native plugin marketplace format)
+- A direct Git repo with skills/commands/agents in `.claude/` or at root level
+
+The Marketplace page in the ACP UI shows:
+
+- Browsable catalogs from each source with search and type filters (skill / command / agent / workflow)
```
Gkrumbach07: we should be able to read plugins marketplace and just stabndlone skills liek if i gavce a repo that had a .claude

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+- ACP runner (production)
+- `/load-workflow` skill (local)
+- Claude Code plugin hook (automated local)
+- Workflow builder UI (creation/editing)
+
+---
+
+## Open Questions
+
+1. **Manifest format**: YAML vs JSON? Should we extend `ambient.json` or create a new `workflow.yaml`? The existing `ambient.json` has `name`, `description`, `systemPrompt`, `startupPrompt`, `rubric` — we'd add `sources` and `prompt`.
```
jwm4: > we'd add `sources` and `prompt`

the idea presented earlier in the document is that prompt would replace a bunch of existing capabilities (systemPrompt, startupPrompt, CLAUDE.md).  I think that makes more sense than adding prompt in addition to the things that are already there.

specs/skills-and-workflows.md
jeremyeder: i think so - because I might want to leverage a hosted agent. in that case i might need auth support, too. the easiest example is the internal productpages MCP which requires auth and contains key ground truth for any agent.

specs/skills-and-workflows.md
jeremyeder: ACP is grounded in workflows, I would want this implementation to reflect that.

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+The workflow manifest (`workflow.yaml` or `ambient.json`) is the core deliverable. It defines prompt + skill sources. Multiple consumers read the same format:
+
+- **ACP runner**: clones sources, adds to `add_dirs`, injects prompt
+- **Local script/skill**: clones sources to temp dirs, passes as `--add-dir`
+- **Claude Code plugin**: `SessionStart` hook reads manifest, sets up environment
+
+The format must support:
+- Prompt text (inline or file reference)
+- Skill sources (Git URL + branch + path)
+- Optional: system prompt, startup prompt, rubric
```
jeremyeder: Credentials for authenticated services?

specs/skills-and-workflows.md
jeremyeder: just want to call out explicitly for MCP - must support auth and tls

specs/skills-and-workflows.md
jeremyeder: We can make adapters as the ecosystem develops, or just rewrite them all. I am working under the assumption that there are no sacred cows at this point. However the concept of workflows, that you can automate them and that they offer maximum power to knowledge-workers, that has to be delivered.

specs/skills-and-workflows.md
jeremyeder: Wouldn't you rather see engineers figuring out + adopting evals? Maybe there is some other way we could encourage that.

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+    - name: Tests added
+      weight: 0.3
+    - name: No regressions
+      weight: 0.2
+```
+
+This format is consumed by:
+- ACP runner (production)
+- `/load-workflow` skill (local)
+- Claude Code plugin hook (automated local)
```
jeremyeder: Is there a marketplace that the ACP project provides? (not the top level one that does the inheritance), for ACP itself - like this /load-workflow?

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+
+## Open Questions
+
+1. **Manifest format**: YAML vs JSON? Should we extend `ambient.json` or create a new `workflow.yaml`? The existing `ambient.json` has `name`, `description`, `systemPrompt`, `startupPrompt`, `rubric` — we'd add `sources` and `prompt`.
+
+2. **Skill versioning**: Sources reference branches today. Should we support tags or SHAs for pinning? What happens when a skill source updates — do sessions get the latest on next start?
+
+3. **Plugin format**: Should workflow export produce a Claude Code plugin (`plugin.json`)? Pros: portable, namespaced, versioned. Cons: plugins cache/copy files which breaks the dynamic reference model.
+
+4. **RHAI alignment**: How does this map to RHAIRFE-1370 (Skills Registry)? Our manifest format and marketplace could inform the product's in-cluster registry design.
```
jeremyeder: By default, we should be using the exact components/feature or aiming to converge. If there are decisions that lead down different paths, we have to discuss.

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+
+---
+
+## Installation & Configuration
+
+### Workspace Level
+
+Items installed at the workspace level are stored in the ProjectSettings CR (`spec.installedItems`). These represent the workspace's **library** — what's available, not what's auto-loaded into every session.
+
+When creating a session, users select which installed items to include. The workflow they choose may also pull in its own skill dependencies from the manifest.
```
Gkrumbach07: this i think we need some thinking basically in the installed items we can install
- individual skills, commands, agents
- plugins
- workflows

at the end of the dya these are just git repo branch and path and we can know that that that path it is one or the other depending on its contents which are defined by what claude code expects

then saved/installed/registered ones can be added adhocx to a session. you can add multiple workflows, multiple skills/commands/agents or even add multiple plugins

specs/skills-and-workflows.md
jeremyeder: Let me explain - the product may not prescribe an agent framework, even though the ACP project may need one. In that case, we pick our solution and ensure interop with RHOAI.

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+
+### Session Level
+
+Skills can be added to a running session via the context panel:
+
+- "Import Skills" in the Add Context dropdown
+- Provide a Git URL + optional branch + path
+- Backend clones, scans, writes skill files to `/workspace/file-uploads/.claude/`
```
Gkrumbach07, line 146: not here, that was temp it should go in where claude expects

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+
+1. **Manifest format**: YAML vs JSON? Should we extend `ambient.json` or create a new `workflow.yaml`? The existing `ambient.json` has `name`, `description`, `systemPrompt`, `startupPrompt`, `rubric` — we'd add `sources` and `prompt`.
+
+2. **Skill versioning**: Sources reference branches today. Should we support tags or SHAs for pinning? What happens when a skill source updates — do sessions get the latest on next start?
+
+3. **Plugin format**: Should workflow export produce a Claude Code plugin (`plugin.json`)? Pros: portable, namespaced, versioned. Cons: plugins cache/copy files which breaks the dynamic reference model.
+
+4. **RHAI alignment**: How does this map to RHAIRFE-1370 (Skills Registry)? Our manifest format and marketplace could inform the product's in-cluster registry design.
+
+5. **Security**: How do we verify skill sources haven't been tampered with? Git commit SHAs provide content-addressable verification. Enterprise customers may need signed manifests.
```
jeremyeder: I don't see these things any differently than traditional source code, so we should anticipate (unsure on timeline) all the expected signing and distribution security.

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+A UI for composing workflows from standalone skills:
+
+- Select skills from the workspace library or browse marketplace
+- Each skill is a reference (source URL + path), not a copy
+- Define the workflow prompt (methodology, phases, instructions)
+- Optionally configure: system prompt, startup prompt, rubric
+- Save as a workflow manifest that can be:
+  - Stored in the workspace
+  - Exported as a Git repo
+  - Exported as a Claude Code plugin (`plugin.json`)
```
Gkrumbach07: it is not mvp so rmeove. what we should have is the abailty to visualze what is ina. plugin/skill/workflow

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+1. User picks a workflow (or "General chat" for none)
+2. The workflow manifest declares its skill dependencies — those are auto-loaded
+3. User can optionally add standalone skills from the workspace library
+4. The session CRD stores the workflow reference + any additional skill sources
+
+This means:
+- Installing 50 skills to the workspace doesn't bloat every session
+- The workflow controls its own dependencies
+- Users can augment with extras per session
+- Workspace-level "always-on" skills could be supported via a flag but are not the default
```
Gkrumbach07, line 179: you should be able to add as manyas you want of workflow, skills, commands, plugins... as long as they are in your registred stuff

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+
+2. **Skill versioning**: Sources reference branches today. Should we support tags or SHAs for pinning? What happens when a skill source updates — do sessions get the latest on next start?
+
+3. **Plugin format**: Should workflow export produce a Claude Code plugin (`plugin.json`)? Pros: portable, namespaced, versioned. Cons: plugins cache/copy files which breaks the dynamic reference model.
+
+4. **RHAI alignment**: How does this map to RHAIRFE-1370 (Skills Registry)? Our manifest format and marketplace could inform the product's in-cluster registry design.
+
+5. **Security**: How do we verify skill sources haven't been tampered with? Git commit SHAs provide content-addressable verification. Enterprise customers may need signed manifests.
+
+6. **Workspace defaults**: Should some workspace-level items be "always-on" (loaded in every session regardless of workflow)? Example: company-wide coding standards skill. Or should this be handled via org-level Claude Code managed settings?
```
jeremyeder: Ensure this top level enforcement feature exists in this feature. Leave it there, inert/feature-flagged off, until enforcement is requested.

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+- **Uploads** — uploaded files (existing)
+
+Users can add skill sources live (Import Skills button). The backend clones the source, writes files to `/workspace/file-uploads/.claude/`, and Claude picks them up immediately. Users can remove individual skills — the file is deleted and Claude stops seeing it.
+
+### Workflow Metadata
+
+The runner's `/content/workflow-metadata` endpoint returns all discovered skills, commands, and agents from:
+- The active workflow's `.claude/` directory
+- Any additional source directories
+- `/workspace/file-uploads/.claude/` (live imports)
```
Gkrumbach07, line 216: its own spot

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+
+- **Repositories** — Git repos cloned as working directories (existing)
+- **Skills** — imported skills/commands/agents with type badges and source links
+- **Uploads** — uploaded files (existing)
+
+Users can add skill sources live (Import Skills button). The backend clones the source, writes files to `/workspace/file-uploads/.claude/`, and Claude picks them up immediately. Users can remove individual skills — the file is deleted and Claude stops seeing it.
+
+### Workflow Metadata
+
+The runner's `/content/workflow-metadata` endpoint returns all discovered skills, commands, and agents from:
```
Gkrumbach07, line 213: across all add dirs

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+### 1. Manual (no tooling)
+
+```bash
+git clone --depth 1 https://github.com/ambient-code/skills.git /tmp/skills
+git clone --depth 1 https://github.com/opendatahub-io/ai-helpers.git /tmp/ai-helpers
+
+claude \
+  --add-dir /tmp/skills/bugfix \
+  --add-dir /tmp/ai-helpers/helpers
+```
```
Gkrumbach07: this is a given, remove

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+
+---
+
+## Open Questions
+
+1. **Manifest format**: YAML vs JSON? Should we extend `ambient.json` or create a new `workflow.yaml`? The existing `ambient.json` has `name`, `description`, `systemPrompt`, `startupPrompt`, `rubric` — we'd add `sources` and `prompt`.
+
+2. **Skill versioning**: Sources reference branches today. Should we support tags or SHAs for pinning? What happens when a skill source updates — do sessions get the latest on next start?
+
+3. **Plugin format**: Should workflow export produce a Claude Code plugin (`plugin.json`)? Pros: portable, namespaced, versioned. Cons: plugins cache/copy files which breaks the dynamic reference model.
```
jeremyeder: I'm on the fence with dynamic reference model. It seems fine for casual usage. I bet people will ask to pin.

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+/load-workflow https://github.com/ambient-code/workflows/tree/main/workflows/bugfix
+```
+
+The skill instructs Claude to:
+1. Fetch the workflow manifest
+2. Clone each skill source to temp directories
+3. Symlink `.claude/skills/` structures into the project
+4. Apply the workflow prompt
+
+This makes ACP workflows portable — anyone with Claude Code can use them without ACP. The skill ships as part of ai-helpers or as its own standalone skill.
```
Gkrumbach07: this should add things more claude native

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+      "type": "command",
+      "command": "${CLAUDE_PLUGIN_ROOT}/scripts/load-workflow.sh"
+    }]
+  }
+}
+```
+
+The hook script reads a `workflow.yaml` from the project root (if present), clones skill sources, and symlinks them into place. Fully automated, no user action needed.
+
+---
```
Gkrumbach07: rtemove

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+      weight: 0.2
+```
+
+This format is consumed by:
+- ACP runner (production)
+- `/load-workflow` skill (local)
+- Claude Code plugin hook (automated local)
+- Workflow builder UI (creation/editing)
+
+---
```
Gkrumbach07: this is athe ambient.json not yaml

specs/skills-and-workflows.md
Gkrumbach07: json

specs/skills-and-workflows.md
```diff
@@ -0,0 +1,333 @@
...
+
+## Open Questions
+
+1. **Manifest format**: YAML vs JSON? Should we extend `ambient.json` or create a new `workflow.yaml`? The existing `ambient.json` has `name`, `description`, `systemPrompt`, `startupPrompt`, `rubric` — we'd add `sources` and `prompt`.
+
+2. **Skill versioning**: Sources reference branches today. Should we support tags or SHAs for pinning? What happens when a skill source updates — do sessions get the latest on next start?
```
Gkrumbach07: yes

jeremyeder: How about we provide the user with choice here - they would opt-in to "Auto-update Harness" at the project and session level.

specs/skills-and-workflows.md
Gkrumbach07: no, it can, but no.

specs/skills-and-workflows.md
Gkrumbach07: ignore

specs/skills-and-workflows.md
Gkrumbach07: ignore

specs/skills-and-workflows.md
Gkrumbach07: we need to chat about always on

specs/skills-and-workflows.md
jwm4: Yes, I definitely want an automated review/repair loop, and I do think we should encourage it.  I just think the way the industry is going for how to do that is via SKILL.md files.

specs/skills-and-workflows.md
Gkrumbach07: Applied in 36b45586.

specs/skills-and-workflows.md
Gkrumbach07: Applied in 36b45586.

specs/skills-and-workflows.md
Gkrumbach07: Applied in 36b45586.

specs/skills-and-workflows.md
Gkrumbach07: Applied in 36b45586.

specs/skills-and-workflows.md
Gkrumbach07: Applied in 36b45586.

specs/skills-and-workflows.md
Gkrumbach07: Applied in 36b45586.

specs/skills-and-workflows.md
Gkrumbach07: Applied in 36b45586.

specs/skills-and-workflows.md
Gkrumbach07: Applied in 36b45586.

specs/skills-and-workflows.md
Gkrumbach07: Applied — consolidated to CLAUDE.md only. Dropped systemPrompt/startupPrompt from ambient.json. No inline prompt field, CLAUDE.md IS the prompt.

specs/skills-and-workflows.md
Gkrumbach07: Applied — sources now support both formats: structured {url, branch, path} object OR a single URL string that gets auto-parsed.

specs/skills-and-workflows.md
Gkrumbach07: Resolved — keeping 'sources' since it covers skills, commands, agents, plugins, and anything else Claude can discover. No type field, runner auto-detects.

specs/skills-and-workflows.md
Gkrumbach07: Yes — sources is generic. Any Git URL with discoverable Claude Code content (skills, agents, commands, plugins) works.

specs/skills-and-workflows.md
Gkrumbach07: Noted — added that multi-agent orchestration is out of scope for this spec. The Agent concept here is single-persona.

specs/skills-and-workflows.md
Gkrumbach07: Updated — we follow Claude Code's plugin format and the Agent Skills (agentskills.io) open standard as the closest cross-tool spec.

specs/skills-and-workflows.md
Gkrumbach07: Applied — updated Discovery section. Scanner now expects three source types: (1) Claude Code plugins (.claude-plugin/plugin.json), (2) marketplace.json catalogs, (3) standalone repos with .claude/.

specs/skills-and-workflows.md
Gkrumbach07: Keeping rubric as-is for now — existing workflows (bugfix, triage) use it. It's optional and doesn't hurt portability if unused.

specs/skills-and-workflows.md
Gkrumbach07: Applied — scanner supports all three: plugins, marketplace catalogs, and standalone .claude repos.

specs/skills-and-workflows.md
Gkrumbach07: Applied — prompt consolidated to just CLAUDE.md. systemPrompt and startupPrompt dropped from ambient.json.

specs/skills-and-workflows.md
Gkrumbach07: Agreed — sources is generic, includes hosted agents. Auth handled via existing workspace credential system.

specs/skills-and-workflows.md
Gkrumbach07: Spec centers workflows as the primary concept. Skills are atoms, workflows are the user-facing thing.

specs/skills-and-workflows.md
Gkrumbach07: Applied — added section on auth. Private sources use existing workspace credential system (GitHub/GitLab integrations + git credential helper).

specs/skills-and-workflows.md
Gkrumbach07: Applied — noted that MCP sources requiring auth and TLS are handled through workspace integration configuration.

specs/skills-and-workflows.md
Gkrumbach07: Agreed — we follow Claude Code plugin format for now, can adapt as ecosystem develops.

specs/skills-and-workflows.md
Gkrumbach07: Rubric kept as-is. Agree evals matter — the format may evolve but the capability stays.

specs/skills-and-workflows.md
Gkrumbach07: Yes — ACP provides its own marketplace via ConfigMap sources. Behind a feature flag.

specs/skills-and-workflows.md
Gkrumbach07: Applied — we follow Claude Code plugin format as the standard. Agent Skills (agentskills.io) open standard.

specs/skills-and-workflows.md
Gkrumbach07: Resolved — everything is a source reference. No type distinction in the data model. Runner auto-detects.

specs/skills-and-workflows.md
Gkrumbach07: Acknowledged — we aim to converge with product direction.

specs/skills-and-workflows.md
Gkrumbach07: Applied — removed file-uploads as skill storage. Skills go where Claude expects them. Exact path needs further discussion (workspace .claude/ vs add_dirs).

specs/skills-and-workflows.md
Gkrumbach07: Removed from spec scope. Acknowledged as future concern.

specs/skills-and-workflows.md
Gkrumbach07: Applied — workflow builder removed. Added visualization: marketplace detail panel + session context panel both show extracted skills/commands/agents readably.

specs/skills-and-workflows.md
Gkrumbach07: Applied — updated selection model. Registry = available, session picks. Users can add any number of sources. 'Always add' defaults need further design.

specs/skills-and-workflows.md
Gkrumbach07: Removed from spec scope. Marketplace itself is behind feature flag.

specs/skills-and-workflows.md
Gkrumbach07: Applied — skills have their own section in context panel.

specs/skills-and-workflows.md
Gkrumbach07: Applied — metadata scans across all loaded sources.

specs/skills-and-workflows.md
Gkrumbach07: Applied — removed explicit env var mention.

specs/skills-and-workflows.md
Gkrumbach07: Applied — sources support optional sha/tag for pinning. Branch default gives auto-update. User's choice.

specs/skills-and-workflows.md
Gkrumbach07: Applied — local usage reframed around Claude-native discovery patterns.

specs/skills-and-workflows.md
Gkrumbach07: Applied — removed Claude Code plugin hook section for local usage.

specs/skills-and-workflows.md
Gkrumbach07: Applied — all references are ambient.json, no YAML.

specs/skills-and-workflows.md
Gkrumbach07: Applied — JSON throughout.

specs/skills-and-workflows.md
Gkrumbach07: Applied — versioning with branch (auto-update) + optional sha/tag (pinning).

Gkrumbach07: Covered — branch default = auto-update. SHA pinning = manual update. User chooses per source.

specs/skills-and-workflows.md
Gkrumbach07: Applied — removed plugin format export from spec.

specs/skills-and-workflows.md
Gkrumbach07: Removed from spec.

specs/skills-and-workflows.md
Gkrumbach07: Removed from spec.

specs/skills-and-workflows.md
Gkrumbach07: Kept as open question — needs further discussion.

specs/skills-and-workflows.md
Gkrumbach07: Agreed — rubric kept but the format may evolve. The goal (automated review/repair) is right, the mechanism can improve.

Gkrumbach07:
bot should not have merged. continue convo here https://github.com/ambient-code/platform/pull/1120

I havbe addressed all comments here already

PR #1133, opened 2026-04-01
fix: use git add -u in merge script to avoid orphaned gitlinks

### LINKED ISSUE #1131: rebase-main-to-alpha.sh: git add -A stages untracked repos as orphaned gitlinks
```
```

**User Story:** _As a release engineer, I want the alpha merge script to only stage tracked-file changes, so that stray directories in the worktree don't create orphaned gitlinks._

## Problem

`scripts/rebase-main-to-alpha.sh` uses `git add -A` on lines 67 and 87 when staging conflict-resolution results during an automated main→alpha merge. `git add -A` stages *all* untracked content in the worktree, including directories that happen to be separate git repositories. Git records these as `160000` gitlink entries (submodule pointers) without a `.gitmodules` entry.

This is the root cause of #1130 — an orphaned `platform-api-server` gitlink that broke all ArgoCD sector deployments.

## Fix

Replace `git add -A` with `git add -u` on lines 67 and 87 of `scripts/rebase-main-to-alpha.sh`.

`git add -u` only stages changes to already-tracked files, so stray directories in the worktree won't be picked up. The script's intent is to stage conflict-resolved files, not to add new untracked content.

ktdreyer:
## Summary

- Replace `git add -A` with `git add -u` in `scripts/rebase-main-to-alpha.sh` (both occurrences on lines 67 and 87)
- `git add -A` stages all untracked content, including stray cloned repositories in the worktree, which get recorded as orphaned gitlinks that break ArgoCD sync
- `git add -u` only stages changes to already-tracked files, which is the correct behavior during conflict resolution

Fixes #1131


ktdreyer:
> Auto reviews are disabled on base/target branches other than the default branch.

Filed https://github.com/ambient-code/platform/issues/1134 to improve that.

PR #1145, opened 2026-04-01
feat: add CodeRabbit integration for AI-powered code review

jeremyeder:
## Summary

- Add CodeRabbit as a native integration in ACP, enabling AI-powered code review inside agentic sessions
- Users store a CodeRabbit API key via the integrations page; the runner injects it as `CODERABBIT_API_KEY` so the `coderabbit` CLI can run local reviews
- Add `.coderabbit.yaml` config and pre-commit hook for enforcing reviews at commit time

### Backend
- CodeRabbit auth handlers: connect, disconnect, status, test endpoints
- Per-user K8s Secret credential storage (`coderabbit-credentials`)
- API key validation against CodeRabbit health endpoint
- Runtime credential fetch for sessions (with RBAC)
- CodeRabbit included in unified integrations status

### Frontend
- CodeRabbit integration card on Integrations page (API key connect/disconnect)
- API client, React Query hooks, Next.js proxy routes
- CodeRabbit shown in session settings integrations panel

### Runner
- Fetches CodeRabbit API key at session startup, sets `CODERABBIT_API_KEY` env var

### Pre-commit / Config
- `.coderabbit.yaml` with project-specific review rules, path instructions, custom pre-merge checks
- `coderabbit-review.sh` pre-commit hook with rate limit handling and 5-minute timeout

### Tests
- Backend: 7 Ginkgo tests for CodeRabbit auth handlers (connect, status, disconnect, validation, RBAC)
- Frontend: integrations panel test updated for CodeRabbit

## Test plan
- [x] Backend builds cleanly (`go build ./...`)
- [x] Backend tests pass (427/427 including 7 new CodeRabbit tests)
- [x] Frontend builds cleanly (`npm run build`)
- [x] Frontend tests pass (587/587)
- [x] CodeRabbit review: no findings
- [ ] Manual: connect CodeRabbit API key via integrations page
- [ ] Manual: verify `CODERABBIT_API_KEY` injected in session pod
- [ ] Manual: verify pre-commit hook runs `coderabbit review` on commit

## Follow-up (separate PRs)
- Default `pre-commit install` to enabled when CodeRabbit integration is configured
- Add `coderabbit` binary to runner Dockerfile

## Summary by CodeRabbit

* **New Features**
  * CodeRabbit integration: connect/disconnect/test endpoints, frontend card, and integration status
  * Runtime credential retrieval for sessions and ambient runner support to inject API key

* **Tests**
  * Added backend tests covering connect/status/disconnect flows

* **Chores**
  * Added pre-commit hook to run CodeRabbit review on staged changes

* **Documentation**
  * Added CodeRabbit integration design and implementation plan


markturansky (commented):
# 🐰 CodeRabbit Integration Review

**tl;dr**: This is a *chef's kiss* implementation of a full-stack integration. Clean patterns, comprehensive tests, thoughtful error handling, and excellent documentation. A few notes below, but this is production-ready.

---

## The Good Stuff ✨

### 1. **Architecture is 💯**
The flow is beautifully simple: API key → K8s Secret → Runtime injection → Pre-commit hook. No overengineering, no unnecessary abstraction. The credential lifecycle follows the exact same pattern as GitHub/GitLab, which means future devs won't need to learn new patterns.

### 2. **Security-conscious design**
- API key validation *before* storage (line 54 in `coderabbit_auth.go`)
- Proper RBAC checks on runtime credential fetch
- Per-user secret isolation (no cross-contamination)
- Conflict retry logic (3 attempts) on K8s Secret updates
- `writeOnly: true` semantics implied by one-way storage

### 3. **Graceful degradation everywhere**
The pre-commit hook is a masterclass in "fail open when appropriate":
```bash
# No CLI? Skip gracefully
# No auth? Skip gracefully  
# Rate limit? Warn, don't block
# Network error? Warn, don't block
# Actual findings? BLOCK 🚫
```
This is the right balance between "enforce quality" and "don't break people's workflows when external services hiccup."

### 4. **Test coverage is excellent**
7 new Ginkgo tests covering:
- Valid/invalid API keys
- Connect/disconnect flows
- Per-user isolation
- RBAC enforcement on runtime fetch
- Frontend integration panel updated

The test mocking pattern (replacing `ValidateCodeRabbitAPIKey` with a test stub) is clean and avoids hitting real APIs.

### 5. **Documentation is A+**
Both the design spec and implementation plan are detailed, opinionated, and actionable. The "Non-Goals" section is especially valuable — it sets clear boundaries and prevents scope creep.

---

## Questions & Observations 🤔

### 1. **Pre-commit hook timeout: 5 minutes**
```bash
OUTPUT=$(timeout 300 "$CR" review --type uncommitted --prompt-only 2>&1)
```
5 minutes is generous! Is this based on observed CodeRabbit CLI behavior on large diffs? Just curious if we've seen it take that long in practice, or if this is "better safe than sorry."

**Why it matters:** Long-running pre-commit hooks can frustrate developers. If CodeRabbit typically finishes in <30s, we might want to tune this down and rely on the transient error handling.

### 2. **API key validation endpoint**
```go
req.Header.Set("Authorization", "Bearer "+apiKey)
resp, err := client.Do(req)
return resp.StatusCode == http.StatusOK, nil
```
The validation hits `/api/v1/health` with the API key. Does CodeRabbit's health endpoint actually validate the key, or does it just return 200 regardless? If it's the latter, we might want to use a different endpoint (like `/api/v1/user` or similar) that requires valid auth.

**Why it matters:** If the health endpoint doesn't check the key, users could "successfully connect" with invalid keys, then get cryptic failures in sessions.

### 3. **Secret storage: cluster-scoped, all users in one Secret**
```go
const secretName = "coderabbit-credentials"
secret.Data[creds.UserID] = b
```
This stores all users' keys in a single K8s Secret, keyed by `userID`. This is consistent with the existing pattern (good!), but as the user base grows, this Secret could get large.

**Potential future optimization:** If you hit hundreds of users, consider sharding (e.g., `coderabbit-credentials-{hash(userID) % 10}`) or switching to per-user Secrets. Not urgent, but worth noting for scale planning.

### 4. **Runner credential fetch: no caching**
The runner fetches CodeRabbit credentials at session startup via:
```python
coderabbit_creds = await fetch_coderabbit_credentials(context)
```
This is synchronous with `populate_runtime_credentials()`, which already fetches GitHub, GitLab, Jira, Google concurrently. Perfect!

But if the credential changes mid-session (user disconnects/reconnects), the session won't pick it up until restart. Is this intentional?

**Why it matters:** Probably fine — credentials rarely change mid-session, and forcing a restart is reasonable. Just wanted to confirm this is the intended behavior.

### 5. **Pre-commit hook: `always_run: true`**
```yaml
- id: coderabbit-review
  always_run: true
  pass_filenames: false
```
This runs the hook even when there are no staged files, but the script itself short-circuits:
```bash
if git diff --cached --quiet; then
    exit 0
fi
```
Why not set `always_run: false` in the config and remove the check from the script? Minor, but curious about the reasoning.

### 6. **Error handling: distinguishing transient vs. permanent failures**
```bash
if echo "$OUTPUT" | grep -qi "rate limit\|network\|timeout\|connection"; then
    echo "CodeRabbit: transient error (see above) — not blocking commit"
    exit 0
fi
```
This grep pattern is clever, but it's a heuristic. If CodeRabbit's error messages change, this could miss transient errors (and block commits unnecessarily) or catch false positives (and let through failures that should block).

**Suggestion:** Consider having the CodeRabbit CLI return specific exit codes for different failure types (e.g., `2` for transient, `1` for findings, `0` for success). Then you can switch on `$EXIT_CODE` instead of grepping output. (This might require a CodeRabbit CLI feature request.)

---

## Nitpicks 🔬

### 1. **Naming inconsistency: credentials vs. credential**
```python
async def fetch_coderabbit_credentials(context: RunnerContext) -> dict:
```
vs.
```python
async def fetch_gitlab_token(context: RunnerContext) -> str:
```
GitLab returns a single token (singular), CodeRabbit returns credentials (plural, but it's just an API key). Not a blocker, just noticed the pattern variance. Might be cleaner to call it `fetch_coderabbit_api_key` or `fetch_coderabbit_credential` (singular).

### 2. **Unused return value in validation**
```go
valid, err := ValidateCodeRabbitAPIKey(c.Request.Context(), req.APIKey)
```
The function signature returns `(bool, error)`, but in `ConnectCodeRabbit` you only check `err`:
```go
if err := ValidateCodeRabbitAPIKey(c.Request.Context(), req.APIKey); err != nil {
```
This works because the function returns `(false, err)` on failure, but the `bool` return is redundant if you're only checking `err`. Consider simplifying to `func ValidateCodeRabbitAPIKey(ctx, key) error`.

### 3. **Frontend: API client typing**
```typescript
export async function connectCodeRabbit(data: CodeRabbitConnectRequest): Promise<void>
```
The API returns `{"message": "CodeRabbit connected successfully"}`, but the TS return type is `void`. Should probably be `Promise<{message: string}>` if you want type safety on the response. (Though if you're not using the response, `void` is fine.)

---

## The Verdict 🎯

This PR is **ready to ship**. The architecture is sound, the error handling is thoughtful, the tests are comprehensive, and the docs are excellent. The questions above are mostly curiosity + future-proofing notes — none are blockers.

**What I especially appreciate:**
- You didn't overcomplicate it. This could have been 5x larger with "features" nobody asked for.
- The pre-commit hook fail-open logic is production-tested wisdom, not naive optimism.
- The implementation plan reads like a cookbook — anyone could follow it.
- Consistent patterns with existing integrations = low cognitive overhead.

**Ship it!** 🚢🐰

---

**P.S.** The irony of CodeRabbit reviewing its own integration PR is not lost on me. I see CodeRabbit left a comment too — curious if it found anything! 😄

jeremyeder:
@ambient-code

jeremyeder:
Superseded by #1315 (rebased onto main with #1307, replaced pre-commit hook with command-type review gate).

PR #1191, opened 2026-04-03
feat: add stopOnRunFinished + improve activity tracking

Gkrumbach07:
## Summary

Adds `stopOnRunFinished` CRD field and improves inactivity detection. Builds on #1180.

### `stopOnRunFinished` (new CRD field)
- When `true`, the backend auto-stops the session on `RUN_FINISHED` event
- Uses in-memory cache (`sync.Map`) to avoid k8s API call on every `RUN_FINISHED` for sessions without the flag
- Wired through: CRD → backend types → session create handler → AG-UI proxy

### Activity tracking improvements
- All AG-UI events now reset the inactivity timer (was only 4 event types — `RUN_STARTED`, `TEXT_MESSAGE_START`, `TEXT_MESSAGE_CONTENT`, `TOOL_CALL_START`)
- Prevents false inactivity detection during long tool calls that don't emit those specific events
- Activity debounce reduced from 60s to 10s to support shorter inactivity timeouts

### Amber GHA updates
- Use `stop-on-run-finished: 'true'` with `timeout: '0'` (no inactivity timeout)
- Fix/custom prompt split: `@amber` alone → fix prompt, `@amber <text>` → custom
- Shell injection fix: comment body via env var
- Session summary: use `||` fallback instead of string concatenation

## Test plan

- [ ] Create session with `stopOnRunFinished: true` — verify it stops on `RUN_FINISHED`
- [ ] Create session without the flag — verify no change in behavior
- [ ] Verify activity tracking keeps sessions alive during long tool calls
- [ ] Test `@amber` comment on a PR — verify fix prompt runs
- [ ] Test `@amber do something` on an issue — verify custom prompt runs

## Release Notes

* **New Features**
  * Sessions can now be configured to automatically stop when agents complete their run.

* **Improvements**
  * Enhanced activity tracking responsiveness with faster status updates.
  * Updated session lifecycle management behavior for improved resource handling.

### DIFF
--- a/.github/workflows/amber-issue-handler.yml
+++ b/.github/workflows/amber-issue-handler.yml
@@ -95,7 +95,8 @@ jobs:
             [{"url": "https://github.com/${{ github.repository }}", "branch": "main"}]
           model: claude-opus-4-6
           wait: 'true'
-          timeout: '60'
+          timeout: '0'
+          stop-on-run-finished: 'true'
 
       - name: Post-session labels and comment
         if: steps.existing.outputs.skip != 'true'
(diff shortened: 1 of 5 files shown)


Gkrumbach07:
Superseded by #1192 (backend) and #1193 (GHA).

PR #1213, opened 2026-04-04
feat(control-plane): CP /token endpoint for runner gRPC auth RHOAIENG-56711

markturansky:
## Summary

- **Eliminates async BOT_TOKEN Secret push/refresh loop** — replaces with synchronous pull from a new CP HTTP endpoint
- **New `internal/tokenserver` package** — `GET /token` validates the caller's K8s SA token via TokenReview, checks `system:serviceaccount:*:session-*-sa` pattern, returns a fresh API token via `OIDCTokenProvider`
- **Runner updated** — `_fetch_token_from_cp()` reads pod SA token from standard K8s mount, calls CP `/token` on startup and every gRPC reconnect; falls back to `BOT_TOKEN` env var for local dev when `AMBIENT_CP_TOKEN_URL` unset
- **`kube_reconciler` cleaned up** — removes `ensureSecret`, `StartTokenRefreshLoop`, `refreshRunnerToken`, `refreshAllRunningTokens`; sets `automountServiceAccountToken: true`; injects `AMBIENT_CP_TOKEN_URL` instead of `BOT_TOKEN` secret ref

## Design rationale

CP is the thing rotating the secret — if it's down for `/token` it's down for cycling too. Synchronous pull eliminates the 3-way race (CP ticker → kubelet propagation → runner read) and removes a class of stale-token failures.

Spec: `docs/internal/design/control-plane.spec.md`

## Test plan

- [ ] Deploy CP with `CP_TOKEN_LISTEN_ADDR=:8080` and `CP_TOKEN_URL=http://<cp-svc>:8080/token`
- [ ] Start a session; confirm runner pod has `automountServiceAccountToken: true` and `AMBIENT_CP_TOKEN_URL` env set
- [ ] Confirm runner fetches token on startup (log: `[GRPC CLIENT] Fetched fresh API token from CP token endpoint`)
- [ ] Force gRPC reconnect; confirm fresh token fetched via CP endpoint
- [ ] Verify `GET /healthz` on token server returns 200
- [ ] Confirm non-runner SA token is rejected (403)
- [ ] Local dev: unset `AMBIENT_CP_TOKEN_URL`, set `BOT_TOKEN`; confirm fallback path works

## Summary by CodeRabbit

* **New Features**
  * Control plane exposes a new /token endpoint that issues API tokens on demand.
  * Runners now obtain tokens from the control plane (AMBIENT_CP_TOKEN_URL) using their ServiceAccount tokens; BOT_TOKEN secret is no longer used.
  * Control plane can optionally ensure a NetworkPolicy to allow API-server access between namespaces.

* **Configuration**
  * Added CP token settings (CP_TOKEN_LISTEN_ADDR, CP_TOKEN_URL) to control plane config.

* **Documentation**
  * Design doc updated to describe the CP token endpoint, runner auth flow, and security implications.

### DIFF
--- a/components/ambient-control-plane/cmd/ambient-control-plane/main.go
+++ b/components/ambient-control-plane/cmd/ambient-control-plane/main.go
@@ -182,20 +182,25 @@ func runKubeMode(ctx context.Context, cfg *config.ControlPlaneConfig) error {
 		inf.RegisterHandler("sessions", sessionRec.Reconcile)
 	}
 
-	errCh := make(chan error, 1)
+	tsErrCh := make(chan error, 1)
 	go func() {
-		if err := startTokenServer(ctx, cfg, tokenProvider); err != nil {
-			errCh <- err
-		}
+		tsErrCh <- startTokenServer(ctx, cfg, tokenProvider)
 	}()
 
-	infErr := inf.Run(ctx)
+	infErrCh := make(chan error, 1)
+	go func() {
+		infErrCh <- inf.Run(ctx)
+	}()
+
 	select {
-	case tsErr := <-errCh:
-		return fmt.Errorf("token server: %w", tsErr)
-	default:
+	case tsErr := <-tsErrCh:
+		if tsErr != nil {
+			return fmt.Errorf("token server: %w", tsErr)
+		}
+		return <-infErrCh
+	case infErr := <-infErrCh:
+		return infErr
 	}
-	return infErr
 }
 
 func startTokenServer(ctx context.Context, cfg *config.ControlPlaneConfig, tokenProvider auth.TokenProvider) error {
--- a/components/ambient-control-plane/internal/tokenserver/handler.go
+++ b/components/ambient-control-plane/internal/tokenserver/handler.go
@@ -23,8 +23,7 @@ const (
 )
 
 type tokenResponse struct {
-	Token     string `json:"token"`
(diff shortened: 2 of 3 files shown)


markturansky:
## Follow-up thoughts

After diving deeper into this change, I want to highlight what makes this PR particularly elegant from a systems design perspective:

**The core insight**: Token propagation delays aren't a bug to be optimized away — they're inherent to the Secret-push model. By flipping to synchronous pull, you've eliminated an entire class of failure (race conditions) rather than trying to tune timeouts and retry logic around it.

**However**, there's one critical gap that could bite in production: **the runner has no retry logic when calling the CP token endpoint** (`_grpc_client.py:832-841`). If the CP is mid-rolling-update during a runner reconnect, the fetch fails immediately and the runner enters an error loop.

Since the entire value prop here is reliability (eliminating races), it feels inconsistent to introduce a new single-point-of-failure without retry. A simple exponential backoff (3 attempts, 1s/2s/4s) would make this bulletproof for transient CP unavailability during deployments.

**Minor question**: The MCP sidecar gets `AMBIENT_CP_TOKEN_URL` injected but I don't see MCP sidecar code changes in this PR. Does it already support this pattern, or is that a follow-up PR?

Everything else looks production-ready. The TokenReview validation, PKCE challenge, graceful shutdown — all the hard parts are done correctly. Just needs that retry safety net.

PR #1308, opened 2026-04-14
feat: operator OOM reproducer for scalability triage (#1300)

### LINKED ISSUE #1300: Operator scalability: code-level improvements to reduce memory footprint
## Context

While investigating the operator OOMKill on vteam-uat (see linked PR), I audited the operator codebase for memory and scalability issues. The immediate fix is bumping resource limits, but there are several code-level improvements that would reduce the operator's memory footprint and improve its behavior at scale.

**Current cluster scale (vteam-uat):** 4,319 AgenticSessions, 674 ProjectSettings, 763 namespaces, 477 pods (80 runner pods). Raw JSON for all sessions: ~25MB, estimated ~75MB+ in Go heap as unstructured maps.

Items are sorted by expected impact, with real metrics from the vteam-uat investigation.

---

## 1. Scope controller-runtime cache to managed namespaces only

**Impact: HIGH — largest single memory reduction**

**Current behavior:** The controller-runtime cache watches ALL AgenticSessions and ALL Pods across all 763 namespaces. The pod watch caches every pod in the cluster (477 pods), even though only 80 are runner pods. The cache predicates only filter which events trigger reconciliation — they do NOT filter what's stored in the cache.

**Measured impact:** The full AgenticSession cache is ~25MB raw JSON (~75MB+ in Go heap). All 477 pods are cached even though 397 are irrelevant (non-runner system pods).

**Fix:** Use `cache.Options.ByObject` with label selectors or namespace selectors to restrict what the controller-runtime cache stores:
- Add a label selector on Pods so only `-runner` pods are cached
- Consider namespace-scoped caching for managed namespaces only (requires `cache.Options.DefaultNamespaces` or per-object overrides)

**Expected improvement:** ~50% reduction in cache memory by excluding non-runner pods and unmanaged namespace resources.

**Files:** `components/operator/internal/controller/agenticsession_controller.go`, `components/operator/main.go` (manager Options)

---

## 2. Unbounded conditions array growth on AgenticSession status

**Impact: HIGH — linear memory growth per session over time**

**Current behavior:** `setCondition()` in `helpers.go` (line 220-258) appends new conditions to the status without any size limit or deduplication. Each reconciliation can add multiple conditions (pod scheduled, pod created, runner started, etc.). Over the lifetime of a session, the conditions array grows unbounded.

**Measured impact:** With 4,319 sessions being reconciled, each carrying potentially hundreds of condition entries, this bloats both the CR storage in etcd and the in-memory cache. A single session with many reconciliation cycles could have a conditions array in the tens of KB.

**Fix:** Implement condition pruning — keep only the most recent N conditions per type (e.g., last 20), or prune conditions older than a threshold. Standard Kubernetes condition semantics keep one entry per type and update `lastTransitionTime`.

**Expected improvement:** Reduces both etcd storage and in-memory cache size per session. At scale, this could reclaim 10-30% of session object memory.

**Files:** `components/operator/internal/handlers/helpers.go` (lines 220-258)

---

## 3. HTTP client created per API call in reconciliation loops

**Impact: MEDIUM — connection pool waste and TIME_WAIT accumulation**

**Current behavior:** In `sessions.go` (lines ~1618, 1650, 1679, 1750), a new `http.Client` is created for each API call inside loops (e.g., iterating over repos to add/remove). Each client creates its own connection pool and TCP connections.

**Measured impact:** During reconciliation of sessions with multiple repos, this creates many short-lived HTTP clients. TCP connections enter TIME_WAIT state and accumulate, consuming file descriptors and memory. With 10 concurrent reconciliations processing repo operations simultaneously, this creates bursts of unnecessary connections.

**Fix:** Create a single package-level `http.Client` with appropriate timeouts and connection pooling. Reuse across all API calls.

**Expected improvement:** Reduces transient memory spikes during reconciliation, eliminates TIME_WAIT socket accumulation. Modest steady-state improvement but significant during burst operations.

**Files:** `components/operator/internal/handlers/sessions.go`

---

## 4. Unbounded project timeout cache

**Impact: MEDIUM — slow linear growth with namespace count**

**Current behavior:** `psTimeoutCache` in `inactivity.go` (lines 57-102) stores timeout entries per namespace in a map with no eviction policy. Every unique namespace accessed adds a permanent entry. The cache has no maximum size.

**Measured impact:** With 763 namespaces, each entry is small (namespace string + timeout value + timestamp), so total size is modest (~100KB). But there's no cleanup — entries for deleted namespaces are never removed.

**Fix:** Add TTL-based eviction (remove entries unused for >1 hour) or switch to an LRU cache with a size cap (e.g., 200 entries). The existing 60-second TTL on individual entries helps, but entries themselves are never removed from the map.

**Expected improvement:** Prevents unbounded map growth. Small absolute impact at current scale but prevents accumulation over months of operation.

**Files:** `components/operator/internal/handlers/inactivity.go` (lines 57-102)

---

## 5. Watch loops without cancellation context

**Impact: MEDIUM — resource leak on shutdown/restart**

**Current behavior:** `WatchNamespaces()` and `WatchProjectSettings()` in `main.go` (lines 160-161) are started as goroutines with `context.TODO()`. They use infinite `for` loops and cannot be gracefully shut down. When the operator is terminated, these goroutines and their associated Kubernetes API watch connections are not cleanly closed.

**Measured impact:** On each of the 128 restarts, old watch connections may not be properly closed before the process is killed. The 2-second sleep between reconnects means the operator could have stale connections during the brief window before OOMKill.

**Fix:** Accept a cancellable context from `main()` (derived from `ctrl.SetupSignalHandler()`), pass it to watch functions, and handle context cancellation for clean shutdown. Better yet, migrate these to controller-runtime controllers (the code comment on line 158-159 already notes this as future work).

**Expected improvement:** Clean shutdown, no resource leaks across restarts. Eliminates a class of subtle bugs under frequent restart scenarios.

**Files:** `components/operator/main.go` (lines 160-161), `components/operator/internal/handlers/namespaces.go`, `components/operator/internal/handlers/projectsettings.go`

---

## 6. Unbounded response body reads

**Impact: LOW — only triggers on error paths**

**Current behavior:** In `sessions.go` (line 1786), `io.ReadAll(resp.Body)` reads the entire HTTP response body into memory without any size limit. If a runner returns a large error response, the entire body is loaded.

**Fix:** Use `io.LimitedReader` with a cap (e.g., 64KB) to prevent unbounded reads.

**Files:** `components/operator/internal/handlers/sessions.go`

---

## Related

- PR for immediate resource limit fix: (linked)

jeremyeder:
## Summary

Reproducer script and pprof support for issue #1300 (operator scalability improvements). Deliberately undersizes the operator's memory limit on a local kind cluster and creates bulk resources until OOMKill, proving the scalability issues at smaller scale.

**Changes:**
- **`components/operator/main.go`** — add optional pprof server on `:6060` (enabled via `ENABLE_PPROF=true`). Zero production impact — opt-in only.
- **`scripts/scalability/reproduce-oom.sh`** — reproducer script that:
  - Patches operator to 128Mi (configurable) memory limit
  - Creates namespaces with `ambient-code.io/managed=true` (triggers ProjectSettings + RBAC cascade)
  - Creates AgenticSession CRs to fill the controller-runtime cache
  - Monitors memory usage and detects OOMKill
  - Optionally captures pprof heap profiles between batches

**Usage:**
```bash
# After make kind-up LOCAL_IMAGES=true CONTAINER_ENGINE=docker
./scripts/scalability/reproduce-oom.sh \
    --mem-limit 128Mi \
    --namespaces 300 \
    --sessions-per-ns 5 \
    --batch-size 50 \
    --pprof \
    --cleanup
```

**Heap profile analysis:**
```bash
go tool pprof pprof-dumps/*/heap-batch-*.pb.gz
```

## Test plan

- [ ] `make kind-up LOCAL_IMAGES=true CONTAINER_ENGINE=docker`
- [ ] Run reproducer with default settings — verify OOMKill detected
- [ ] Run with `--pprof` — verify heap profiles captured
- [ ] Run with `--cleanup` — verify namespaces deleted and operator restored
- [ ] Verify pprof profiles show controller-runtime cache as top allocator

Closes #1300 (reproducer only — code fixes are separate PRs)

## Summary by CodeRabbit

* **New Features**
  * Optional runtime profiling HTTP endpoint, enabled via environment toggle.

* **Chores**
  * Added a scalability test script to reproduce OOM conditions with configurable limits, batching, optional heap profiling, and cleanup.
  * Ignore generated scalability artifacts and profiling dumps in version control.

* **Performance & Reliability**
  * Controller caching narrowed to relevant pods and prunes large object payloads.
  * Reused HTTP client and limited error-body reads to reduce overhead.

* **Bug Fixes**
  * Bounded inactivity-timeout cache with eviction to prevent unbounded growth.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -118,6 +118,10 @@ e2e/langfuse/.env.langfuse-keys
 logs/
 reports/
 
+# Scalability test output
+pprof-dumps/
+scalability-runs/
+
 # Security scan artifacts
 .security-scan/
 .security-scan.zip
--- a/components/operator/internal/handlers/inactivity.go
+++ b/components/operator/internal/handlers/inactivity.go
@@ -53,6 +53,10 @@ func init() {
 
 // --- Project-level timeout cache ---
 
+// maxTimeoutCacheEntries caps the namespace timeout cache size to prevent
+// unbounded growth in clusters with many namespaces.
+const maxTimeoutCacheEntries = 500
+
 // projectTimeoutCache caches inactivityTimeoutSeconds from ProjectSettings per namespace.
 type projectTimeoutCache struct {
 	mu      sync.Mutex
@@ -93,8 +97,28 @@ func getProjectInactivityTimeout(namespace string) int64 {
 		}
 	}
 
-	// Update cache under lock
+	// Update cache under lock, evicting stale entries if over capacity
 	psTimeoutCache.mu.Lock()
+	if len(psTimeoutCache.entries) >= maxTimeoutCacheEntries {
+		// Evict expired entries first
+		now := time.Now()
+		for k, v := range psTimeoutCache.entries {
+			if now.Sub(v.fetchedAt) >= inactivityTimeoutCacheTTL {
+				delete(psTimeoutCache.entries, k)
+			}
+		}
+		// If still over capacity, clear the oldest half
+		if len(psTimeoutCache.entries) >= maxTimeoutCacheEntries {
+			count := 0
+			for k := range psTimeoutCache.entries {
+				delete(psTimeoutCache.entries, k)
+				count++
+				if count >= maxTimeoutCacheEntries/2 {
+					break
+				}
+			}
+		}
+	}
 	psTimeoutCache.entries[namespace] = projectTimeoutEntry{timeout: result, fetchedAt: time.Now()}
 	psTimeoutCache.mu.Unlock()
 
--- a/components/operator/internal/handlers/sessions.go
+++ b/components/operator/internal/handlers/sessions.go
@@ -28,6 +28,20 @@ import (
 	"k8s.io/client-go/util/retry"
 )
 
+// Shared HTTP client for runner API calls. Reusing a single client enables
+// connection pooling and avoids TIME_WAIT socket accumulation.
+var runnerHTTPClient = &http.Client{
+	Timeout: 10 * time.Second,
+	Transport: &http.Transport{
+		MaxIdleConnsPerHost: 10,
+		IdleConnTimeout:     30 * time.Second,
+	},
+}
+
+// maxErrorBodyBytes limits how much of an error response body we read into
+// memory. Prevents unbounded allocation if a runner returns a large error page.
+const maxErrorBodyBytes = 64 * 1024 // 64 KB
+
 // handleAgenticSessionEvent is the legacy reconciliation function containing all session
 // lifecycle logic (~2,300 lines). It's called by ReconcilePendingSession() wrapper.
 //
@@ -1615,8 +1629,7 @@ func reconcileSpecReposWithPatch(sessionNamespace, sessionName string, spec map[
 		}
 		req.Header.Set("Content-Type", "application/json")
 
-		client := &http.Client{Timeout: 10 * time.Second}
-		resp, err := client.Do(req)
+		resp, err := runnerHTTPClient.Do(req)
 		if err != nil {
 			log.Printf("[Reconcile] Failed to add repo via runner: %v", err)
 			continue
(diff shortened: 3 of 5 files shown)


jeremyeder:
The reproducer confirmed that these fixes don't help enough to matter. So I'm going to close this for now. The real fix will be establishing quotas at some level within the cluster. the operator currently scales with the # of agenticsession CRs. so its memory grows in that way. There are many solutions - this patch could be one of them. Deferring for now until we hit an actual need. 512MB is a joke - we should have never hit this.

PR #1396, opened 2026-04-21
learned: Session data uses K8s CRDs not PostgreSQL (demo mo8yuc4j)

jeremyeder:
## New Memory

**Type:** correction
**Source:** Manual entry

---

Session data uses K8s CRDs not PostgreSQL. The operator creates Jobs from CRDs. PostgreSQL is only used by the api-server.


jeremyeder:
Test PR — closing.

PR #1471, opened 2026-04-27
fix(api-server): handle Keycloak service-account- prefix in OIDC username matching

markturansky:
## Summary

- Keycloak prefixes service account usernames with `service-account-` in JWT claims, causing the gRPC middleware to fail matching the configured `GRPC_SERVICE_ACCOUNT` username and not assigning `CallerTypeService`.
- Extracts an `isServiceAccount()` helper that accepts both the exact configured name and the Keycloak-prefixed variant (`service-account-<name>`), used in both the unary and stream gRPC interceptors.
- Adds table-driven tests covering exact match, prefixed match, no match, empty inputs, and partial prefix cases.

## Test plan

- [x] `TestIsServiceAccount` passes with all 6 cases
- [ ] Existing `TestBearerTokenAuth` and gRPC interceptor tests continue to pass
- [ ] Deploy to staging and verify service account gRPC calls are correctly identified as `CallerTypeService` with Keycloak-issued tokens


markturansky:
Closing — this fix was already merged to main via #1465.

PR #1475, opened 2026-04-28
spec(api-server): add SessionEvent Kind for durable AG-UI event store

jsell-rh:
Introduces SessionEvent as the canonical append-only record of all AG-UI protocol events in a session, replacing agui-events.jsonl and the overloaded non-user rows that grpc_push_middleware wrote to session_messages.

Key design decisions:
- SessionMessage narrowed to runner inbox (user messages only); matches the existing WatchSessionMessages watcher which discards non-user events
- SessionEvent stores raw AG-UI JSON verbatim; SSE replay pipes payload directly without transformation (fixes the legacy wrapper format bug)
- run_id and thread_id are first-class columns enabling per-run compaction and targeted queries
- Compaction: TEXT_MESSAGE_CONTENT rows marked superseded=true after RUN_FINISHED; MESSAGES_SNAPSHOT appended as canonical transcript
- /agui-events endpoint with since=<seq> (reconnect) and run_id= filters
- grpc_push_middleware must migrate to PushSessionEvent; GRPCMessageWriter is retired

### DIFF
--- a/docs/internal/design/ambient-model.spec.md
+++ b/docs/internal/design/ambient-model.spec.md
@@ -2,7 +2,7 @@
 
 **Date:** 2026-03-20
 **Status:** Proposed — Pending Consensus
-**Last Updated:** 2026-04-28 — added `ScheduledSession` Kind; added session operational sub-resources (workspace, files, git, repos, tasks, runner protocol); added generic proxy surface for backend passthrough; updated coverage matrix: all ScheduledSession commands implemented; session sub-resources (workspace/files/git/repos/operational/runner protocol) implemented in API server; generic proxy plugin implemented
+**Last Updated:** 2026-04-28 — added `ScheduledSession` Kind; added session operational sub-resources (workspace, files, git, repos, tasks, runner protocol); added generic proxy surface for backend passthrough; updated coverage matrix: all ScheduledSession commands implemented; session sub-resources (workspace/files/git/repos/operational/runner protocol) implemented in API server; generic proxy plugin implemented; added `SessionEvent` Kind (durable AG-UI event store replacing `agui-events.jsonl`); clarified `SessionMessage` as runner inbox only; added `/agui-events` SSE endpoint with compaction and `since`/`run_id` filters
 **Guide:** `ambient-model.guide.md` — implementation waves, gap table, build commands, run log
 **Design:** `credentials-session.md` — full Credential Kind design spec and rationale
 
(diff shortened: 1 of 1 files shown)


markturansky (commented):
## Amber Review — PR #1475 (draft)

spec(api-server): SessionEvent Kind for durable AG-UI event store

---

## Summary

A substantial, well-considered spec. The `SessionMessage` narrowing (user inbox only) and `SessionEvent` separation is architecturally sound — the existing `WatchSessionMessages` gRPC watcher already discarded non-user events, so the schema is being brought into alignment with actual usage. The LISTEN/NOTIFY fan-out design, compaction strategy, and implementation checklist are thorough. One correctness issue with compaction atomicity needs to be fixed before implementation starts.

---

## Findings

### Critical

**C1 — Compaction steps are not specified as atomic**

The spec lists three compaction steps:
1. `UPDATE session_events SET superseded = true WHERE ... AND event_type IN (...)`
2. `INSERT` a `MESSAGES_SNAPSHOT` row
3. `INSERT` `STATE_SNAPSHOT` / `ACTIVITY_SNAPSHOT` rows

If step 2 or 3 fails after step 1 completes, delta rows are permanently marked `superseded=true` with no replacement snapshot. SSE replay excludes `superseded=true` rows — that portion of session history becomes permanently inaccessible without manual DB intervention.

All three steps must be a single database transaction:

```
Compaction MUST execute all three steps inside a single database transaction.
If the MESSAGES_SNAPSHOT INSERT fails, the transaction rolls back and streaming
delta rows remain accessible. A retry can safely re-attempt the full compaction.
```

### Major

**M1 — `session_messages` narrowing migration is destructive**

"Existing non-user rows in `session_messages` should be purged or migrated during the cutover migration." Purging rows from a production table is irreversible. The spec should require:
- A point-in-time backup before the migration
- A rollback plan if `CHECK (event_type = 'user')` fails against existing data
- Whether this can run online (with active sessions) or requires a maintenance window

### Minor

**N1 — `GRPCMessageWriter` and `_synthesize_run_error` retirement**

The spec correctly calls out "must be deployed atomically in the same release." Consider making this mechanically enforced (e.g., a startup-time check that `PushSessionEvent` is reachable before `GRPCMessageWriter` is removed from the code path), not just a release convention.

**N2 — `by_alias=True` bug is pre-existing, not only in new code**

The spec notes `_event_to_payload()` currently produces snake_case. This bug affects the current `session_messages` write path too. If SSE consumers already expect camelCase from the runner's live `/events` stream, the DB-backed replay will produce mismatched field names at cutover. Worth a targeted fix and test before `PushSessionEvent` goes live, independent of the rest of this spec.

**N3 — `IS NOT DISTINCT FROM` is PostgreSQL-specific**

Acceptable for this platform, but noting for future portability audits.

---

## What's strong

- Global BIGSERIAL with `(session_id, seq)` index is the right tradeoff — ordered per-session reads without per-session locking
- LISTEN/NOTIFY drain step (discard buffered payloads with `seq ≤ last_replayed_seq`) correctly prevents duplicates at reconnect boundary
- `since=<seq>` cursor is clean and stateless
- 7-step implementation checklist is complete and correctly sequenced
- `superseded` flag over row deletion is the right durability choice
- Calling out that `GRPCMessageWriter` + `_synthesize_run_error` must be retired together prevents split-brain

---

## Before marking ready for review

Fix C1 in the spec text — compaction must be explicitly specified as a single transaction. Address M1 with an operational plan. Everything else can be resolved during implementation.

*— Amber*

PR #1477, opened 2026-04-29
spec(control-plane): add Workspace Initialization for repo cloning

jsell-rh:
## Summary

- Add **Workspace Initialization** section to `control-plane.spec.md` documenting the init container pattern for cloning repos into `/workspace/repos/` before the runner starts
- Documents trigger conditions (`session.RepoURL` or `session.Repos`), repo URL normalization, credential injection for private repos, and the current gap status
- Adds `REPOS_JSON` to the environment variables table

## Context

Sessions created with `--repo-url` via the CLI/TUI currently produce an empty directory at `/workspace/repos/<name>` — the control plane sets `REPOS_JSON` as an env var but does not create an init container to actually clone the repo. The operator implements this correctly; this spec defines the same behavior for the control plane so it can be implemented.

## Test plan

- [ ] Review spec section for completeness against operator implementation
- [ ] Verify init container behavior matches operator pattern in `reconcileSpecReposWithPatch`
- [ ] Confirm credential injection approach is consistent with CP credential resolver

### DIFF
--- a/specs/control-plane/control-plane.spec.md
+++ b/specs/control-plane/control-plane.spec.md
@@ -100,6 +100,15 @@ The CP creates a Pod (not a Job) for each session. Key pod attributes:
 
 Each section is joined with `\n\n`. Empty sections are omitted. If all four are empty, `INITIAL_PROMPT` is not set and the runner waits for a user message via gRPC.
 
+### Workspace Initialization and State Persistence
+
+Workspace setup (repo cloning, S3 state hydration) and ongoing state persistence (S3 sync sidecar) are specified in dedicated specs:
+
+- **[Workspace Initialization](workspace-init.spec.md)** — init container that clones repos, restores S3 state, and prepares `/workspace` before the runner starts
+- **[State Persistence](state-persistence.spec.md)** — sidecar that periodically syncs workspace state to S3 and backs up git repo state
+
+Both are **not yet implemented** in the CP. The operator implements both patterns. Sessions created via the CP currently have ephemeral workspaces with empty repo directories.
+
 ### Environment Variables Injected into Runner Pod
 
 | Var | Value | Purpose |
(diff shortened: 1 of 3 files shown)


markturansky (commented):
## Amber Review — PR #1477 (draft)

spec(control-plane): Workspace Initialization + State Persistence

---

## Summary

Three-file change: a pointer section added to `control-plane.spec.md`, and two full new spec files (`workspace-init.spec.md`, `state-persistence.spec.md`). The PR title undersells the scope — this is effectively two complete specifications, not just a workspace init section. Both specs are well-structured and follow the living-doc convention with clear **Status: Not implemented** sections.

---

## Findings

### Major

**M1 — Scope mismatch between PR title and content**

The title says "add Workspace Initialization for repo cloning" but the PR adds a second full spec (`state-persistence.spec.md`, 187 lines) covering the state-sync sidecar. These are related but distinct features. Consider either:
- Updating the PR title/description to reflect both specs, or
- Splitting state-persistence into a follow-up PR so reviewers know what they're signing off on

Neither blocks merging, but reviewers who see the PR title may miss that they're also approving the persistence design.

### Minor

**N1 — Root privilege exception not flagged for security sign-off**

`workspace-init.spec.md` correctly notes the init container runs as root (for `chown` to UID 1001). This is an intentional exception to the platform's `runAsNonRoot` convention. The spec justifies it inline but doesn't call out that this exception needs explicit platform security approval before implementation. Suggest adding a note like:

```
> **Security exception:** root is required for chown. This must be reviewed
> by the platform security team before the init container is shipped to production.
```

**N2 — `session.Repos` gap called out inline but not tracked**

`workspace-init.spec.md` correctly flags: "Support for `session.Repos` (multi-repo array) MUST be added — it is a field on the session model but is not read by the CP today." This is good. If there's a tracking issue for this gap, linking it here would prevent it from being lost.

**N3 — hydrate.sh update scope**

Both specs reference `hydrate.sh` needing updates (`AMBIENT_CP_TOKEN_URL` support). The Status sections cover this, which is the right place. No action needed — noting for reviewers.

---

## Score

Both specs are solid. M1 is the only thing worth acting on before requesting review from `markturansky`. N1-N3 are documentation quality items.

*— Amber*

PR #1496, opened 2026-05-05
fix(frontend): parse cron scheduled session times as UTC

machadovilaca:
cron-parser defaults to the browser's local timezone, so "Next 3 runs" shows times offset from UTC even though the cron expression is stored and executed as UTC
  
edit

shows 7:30 AM UTC (8:30 AM GMT+1) for next run

list

shows a different time: At 08:30 AM UTC

Fixed by passing { tz: "UTC" } to CronExpressionParser.parse() so next run calculations match UTC

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed cron expression parsing to explicitly use UTC timezone instead of relying on system defaults, ensuring consistent and predictable scheduling behavior across all environments.


jsell-rh:
@machadovilaca Nice catch, thank you!

Do you think it would be better to consistently show times in the browser's timezone? Including making the cron expression entry assume browser's TZ.

Especially for the folks that aren't +/- 1 UTC, the offset calculation can be a point of friction. 

It would require a bit of transformation to take the hour from the cron entry on the frontend so that the user inputs in their local TZ and the API call uses UTC, but it might be a nice quality of life improvement. 

What do you think?

machadovilaca:
> @machadovilaca Nice catch, thank you!
> 
> Do you think it would be better to consistently show times in the browser's timezone? Including making the cron expression entry assume browser's TZ.
> 
> Especially for the folks that aren't +/- 1 UTC, the offset calculation can be a point of friction.
> 
> It would require a bit of transformation to take the hour from the cron entry on the frontend so that the user inputs in their local TZ and the API call uses UTC, but it might be a nice quality of life improvement.
> 
> What do you think?

@jsell-rh i think that makes sense
but maybe something like what's shown on the last run column

```
**Last Run**
9:50 AM UTC (10:50 AM GMT+1)
```

basically 

```
**Schedule (UTC)**
At 08:30 AM
30 8 * * *
```

would become

```
**Schedule**
At 08:30 AM UTC (09:30 AM GMT+1)
30 8 * * *
```

should this be done as part of this PR, or in a follow-up?

jsell-rh:
@machadovilaca, that looks good to me.

I'll approve this workflow to run, and kindly ask that you open a follow-up with the local TZ changes. 

Thank you!

PR #1520, opened 2026-05-07
fix: create parent directory before moving workflow clone

guzalv:
## Summary

Custom workflows loaded via "Load custom" (git URL + branch, no subpath) silently fail — the session starts with an empty workflow directory. None of the workflow's commands, system prompt, or `CLAUDE.md` are available to the agent.

### 1. Missing parent directory for workflow clone

The init container (`hydrate.sh`) and runtime clone (`workflow.py`) both attempt to move the cloned workflow to `/workspace/workflows/<name>`, but the parent directory `/workspace/workflows/` is never created beforehand. The `mkdir -p` call exists in the subpath code path but is missing from the non-subpath path (which is the common case — entire repo = the workflow). The move fails silently (`set +e` in hydrate.sh, `except` in workflow.py), and the runner later creates an empty directory as the CWD.

### 2. Repo and workflow cloning coupled to S3 availability

`hydrate.sh` calls `exit 0` when S3 credentials are missing, which skips **everything** after line 98 — including all git credential setup, repo cloning, and workflow cloning. These operations are independent of S3 state persistence and should always run. This means any deployment without S3 configured (or with temporarily unavailable S3) gets no repos and no workflows.

## Who is affected

- **Missing mkdir**: All users loading a custom workflow without a subpath (the common case). The subpath code path works correctly, which is likely why this wasn't caught.
- **S3 coupling**: Any deployment where S3 credentials are not configured or temporarily unavailable. The platform otherwise functions fine without S3 (sessions start, agent works), but repos and workflows silently don't load.

### `hydrate.sh`
- Add `mkdir -p "$(dirname "$WORKFLOW_FINAL")"` before `mv` in the no-subpath and subpath-fallback branches
- Replace the `exit 0` when S3 is unconfigured with a conditional block: S3-specific operations (rclone setup, state download, repo state restore) are skipped, but git credential setup, repo cloning, and workflow cloning always run

### `workflow.py`
- Add `workflow_final.parent.mkdir(parents=True, exist_ok=True)` before `shutil.move()` in the non-subpath and subpath-fallback branches of `clone_workflow_at_runtime()`

## Test plan
- [ ] Load a custom workflow (URL + branch, no subpath) on a fresh session — verify `/workspace/workflows/<name>/` is populated
- [ ] Load a custom workflow on a deployment **without S3 configured** — verify workflow still clones
- [ ] Load a custom workflow with a valid subpath — verify extraction still works (unchanged code path)
- [ ] Load a custom workflow with a nonexistent subpath — verify fallback to entire repo works
- [ ] Full deployment with S3 configured — verify state hydration still works as before
- [ ] `make kind-up LOCAL_IMAGES=true` with local builds of ambient-runner and state-sync

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced workflow cloning to properly handle edge cases where workflow subpaths are missing, ensuring required directories are created before filesystem operations.
  * Improved workflow repository preparation to conditionally manage operations based on storage availability, preventing unnecessary script termination.

### DIFF
--- a/components/runners/ambient-runner/ambient_runner/endpoints/workflow.py
+++ b/components/runners/ambient-runner/ambient_runner/endpoints/workflow.py
@@ -149,10 +149,12 @@ async def clone_workflow_at_runtime(
                 logger.warning(f"Subpath '{subpath}' not found, using entire repo")
                 if workflow_final.exists():
                     shutil.rmtree(workflow_final)
+                workflow_final.parent.mkdir(parents=True, exist_ok=True)
                 shutil.move(str(temp_dir), str(workflow_final))
         else:
             if workflow_final.exists():
                 shutil.rmtree(workflow_final)
+            workflow_final.parent.mkdir(parents=True, exist_ok=True)
             shutil.move(str(temp_dir), str(workflow_final))
 
         logger.info(f"Workflow '{workflow_name}' ready at {workflow_final}")
--- a/components/runners/state-sync/hydrate.sh
+++ b/components/runners/state-sync/hydrate.sh
@@ -345,6 +345,7 @@ if [ -n "$ACTIVE_WORKFLOW_GIT_URL" ] && [ "$ACTIVE_WORKFLOW_GIT_URL" != "null" ]
             fi
         else
             # No subpath - use entire repo
+            mkdir -p "$(dirname "$WORKFLOW_FINAL")"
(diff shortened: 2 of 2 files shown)


guzalv:
Splitting into two focused PRs: one for mkdir/permissions/security fixes, one for S3 decoupling.

PR #1557, opened 2026-05-11
fix(runner): write Google credentials as {email}.json with timezone-naive expiry

markturansky:
## Summary

- **Strip trailing `Z` from `expiresAt`** before writing to credential file — workspace-mcp's `datetime.fromisoformat()` fails on Python 3.10 with the `Z` suffix, causing it to treat the expiry as `None` and the token as invalid
- **Write credential file as `{email}.json`** instead of `credentials.json` — workspace-mcp expects this naming convention for proper user identification in single-user mode
- **Scan credentials directory for any `.json` file** in `check_mcp_authentication` instead of hardcoding `credentials.json`

Fixes the UAT Google OAuth failure where workspace-mcp rejected valid pre-fetched credentials and fell back to its own inaccessible OAuth flow (PKCE mismatch with backend callback).

## Root Cause

Two issues caused workspace-mcp to reject valid pre-fetched Google OAuth credentials:

1. **Expiry format**: Backend returns `expiresAt` in RFC3339 (`2026-05-11T19:58:05Z`), but workspace-mcp parses it via `datetime.fromisoformat()` which fails on Python 3.10 with the trailing `Z`. workspace-mcp then falls back to its own OAuth flow.

2. **Credential filename**: Runner wrote `credentials.json`, but workspace-mcp's `list_users()` expects `{email}.json`. In single-user mode it finds and loads any `.json` file, but the user identifier becomes `credentials` instead of the actual email.

When workspace-mcp falls back to its own OAuth flow, it generates PKCE `code_challenge` + hex `state`, but the backend's `/oauth2callback` doesn't have the matching `code_verifier`, causing Google to return `"Missing code verifier"`.

## Test plan

- [x] All 681 runner tests pass (0 failures)
- [x] `test_google_oauth_access_token_format` verifies email-based filename and Z-stripped expiry
- [x] `test_mcp_auth.py` Google workspace auth tests updated for directory scanning
- [ ] Deploy to UAT and verify Google Drive works in a new session

## Summary by CodeRabbit

* **New Features**
  * Enhanced Google Workspace MCP authentication to support multiple credential configurations and per-user OAuth token storage.
  * Added support for reading credentials from directory-based structures with fallback options.

* **Bug Fixes**
  * Improved credential file handling and token expiration management with better error logging.

* **Tests**
  * Refactored authentication tests to use real temporary credential directories for more accurate validation.

### DIFF
--- a/components/runners/ambient-runner/ambient_runner/bridges/claude/mcp.py
+++ b/components/runners/ambient-runner/ambient_runner/bridges/claude/mcp.py
@@ -19,6 +19,9 @@ from ambient_runner.platform.utils import get_bot_token
 
 logger = logging.getLogger(__name__)
 
+_WORKSPACE_CREDS_DIR = Path("/workspace/.google_workspace_mcp/credentials")
+_SECRET_CREDS_DIR = Path("/app/.google_workspace_mcp/credentials")
+
 
 DEFAULT_ALLOWED_TOOLS = [
     "Read",
@@ -169,19 +172,16 @@ def log_auth_status(mcp_servers: dict) -> None:
 # ---------------------------------------------------------------------------
 
 
-def _read_google_credentials(
-    workspace_path: Path, secret_path: Path
-) -> dict[str, Any] | None:
-    cred_path = workspace_path if workspace_path.exists() else secret_path
-    if not cred_path.exists():
+def _load_credential_file(path: Path) -> dict[str, Any] | None:
+    if not path.exists():
         return None
     try:
-        if cred_path.stat().st_size == 0:
+        if path.stat().st_size == 0:
             return None
-        with open(cred_path, "r") as f:
+        with open(path, "r") as f:
             return json.load(f)
     except (json.JSONDecodeError, OSError) as e:
-        logger.warning(f"Failed to read Google credentials: {e}")
+        logger.warning(f"Failed to read Google credentials from {path.name}: {e}")
         return None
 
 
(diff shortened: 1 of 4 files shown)


markturansky:
## Review: fix(runner): write Google credentials as {email}.json with timezone-naive expiry

Clean, targeted fix with solid root cause analysis. The three-part fix (expiry format, filename, directory scan) maps directly to the two root causes. Test quality improvement (real filesystem over mocks) is a net positive.

Two minor things worth a look:

---

### 🟡 `_read_google_credentials(json_file, json_file)` — same path passed twice

```python
candidate = _read_google_credentials(json_file, json_file)
```

`_read_google_credentials` is designed as a two-location fallback — try `workspace_path`, fall back to `secret_path`. Passing the same file for both is functionally correct (we found it via `glob`, so it exists, so `workspace_path` always wins), but it misuses the function's interface and will confuse whoever reads this next.

Since we already know the file exists at this point, consider either:
- A single-path helper: `_read_google_credential_file(json_file)`  
- Or `_read_google_credentials(json_file, Path("/dev/null"))` to make the intent explicit

---

### 🟢 No log when scanning multiple `.json` files

The directory scan silently picks the first valid file alphabetically. For single-user mode this is fine, but in practice the credentials dir could accumulate stale files. A one-liner log of which file was chosen would make debugging much easier:

```python
logger.debug("Using Google credentials from %s", json_file)
```

---

### ✅ `Z`-stripping is consistent with `_parse_token_expiry`

Worth noting for reviewers: `_parse_token_expiry` in `mcp.py` already handles timezone-naive strings — if `dt.tzinfo is None` it adds UTC. So writing a naive expiry string is safe for our own read path too. The fix targets workspace-mcp's external parser, but there's no regression risk on our side.

---

### ✅ Test improvements are solid

Replacing the mock-based tests with real `tmp_path` + `monkeypatch` tests means the scanning logic is actually exercised. The `assert written["expiry"] == "2099-01-01T00:00:00"` assertion directly pins the Z-stripping behavior — good regression guard.

LGTM pending the `_read_google_credentials` cleanup.

jsell-rh (commented):
### Minor observations

1. **`_read_google_credentials(json_file, json_file)`** — both args are the same path now. Works because the workspace/secret fallback moved to the outer loop, but the function will try the same file twice on a read failure. Could pass a single-path helper or skip the redundant retry.

2. **`sorted(creds_dir.glob("*.json"))` priority** — if both `credentials.json` and `user@example.com.json` exist, alphabetical sort picks the legacy file first. The write path cleans up the legacy file so this shouldn't happen in practice, but worth noting.

3. **`import shutil` inside `finally`** (`test_shared_session_credentials.py:660`) — move to top-of-file imports.

None of these are blocking.

---
*jsell-rh's Ambient Review Bot \<jsell+ambient-review-bot@redhat.com\>*

PR #1562, opened 2026-05-12
fix(backend): remove initialPrompt skip that blocked display name fallback

quay-devel:
## Summary

- **Removed** the 5-line `initialPrompt` skip guard in `agui_proxy.go:1005-1009` that prevented the AG-UI proxy from acting as a fallback when CreateSession's async display name generation failed
- **Updated** the comment in `sessions.go` to document the new fallback relationship between the two generation paths
- **Added** 4 regression tests for `triggerDisplayNameGenerationIfNeeded` covering the key code paths

## Problem

Sessions intermittently showed raw session IDs (`session-<UUID>`) instead of friendly display names. The display name system has two generation paths:

- **Path A** (CreateSession handler): fires `GenerateDisplayNameAsync` in a goroutine with 10s timeout
- **Path B** (AG-UI proxy): generates on first `/agui/run` POST

Path B intentionally skipped messages matching `initialPrompt` to prevent double-generation. But this also blocked Path B from acting as a fallback when Path A failed (API timeout, Vertex errors, missing credentials). The only recovery was the user sending a second message.

## Fix

The `initialPrompt` skip was **redundant** — two existing guards already prevent double-generation:

1. `ShouldGenerateDisplayName(spec)` returns `false` if `displayName` is already set
2. `updateSessionDisplayNameInternal` has a first-write-wins race guard at `display_name.go:302-307`

Removing the skip lets Path B naturally retry when Path A fails, with no risk of overwrites. Worst case in a race: one redundant Haiku API call (~$0.001).

## Test plan

- [x] All 26 websocket package tests pass (including 4 new regression tests)
- [x] `gofmt` and `go vet` clean
- [ ] Verify in staging: create sessions with `initialPrompt`, confirm display name appears even when Vertex API is slow/unavailable

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved display name generation to be more consistent and reliable when creating sessions with initial prompts.

* **Tests**
  * Added comprehensive test coverage for display name generation scenarios, including edge cases.


markturansky (approved):
LGTM

markturansky:
# Amber Code Review — PR #1562

**PR:** fix(backend): remove initialPrompt skip that blocked display name fallback
**Scope:** `handlers/display_name.go`, `websocket/agui_proxy.go`, `websocket/agui_proxy_test.go`

---

## Summary

Clean, targeted bug fix. The problem analysis in the PR description is accurate: the `initialPrompt` skip was redundant given the two existing guards (`ShouldGenerateDisplayName` + first-write-wins in `updateSessionDisplayNameInternal`). Removing it is the minimal correct change. One test reliability issue needs attention before merge.

---

## Findings

### Blocker

None.

---

### Critical

None.

---

### Major

**1. `TestTriggerDisplayName_SkipsWhenNoUserMessage` cannot actually verify the skip — `agui_proxy_test.go:270-290`**

This test asserts `dn == ""` after calling `triggerDisplayNameGenerationIfNeeded` with only assistant messages. But it does **not** mock `handlers.GenerateDisplayNameAsync`. If the skip logic is broken and the real function is called, it launches a goroutine that tries to reach the LLM API — fails silently in the test environment — and `dn` remains `""` anyway. The test passes whether the skip works or not.

The test for the positive path (`TestTriggerDisplayName_InitialPromptNotSkipped`) correctly uses a `called bool` mock. This test should do the same in the negative:

```go
func TestTriggerDisplayName_SkipsWhenNoUserMessage(t *testing.T) {
    cleanup := setupDisplayNameTest(t, map[string]interface{}{
        "initialPrompt": "Help me debug auth",
    })
    defer cleanup()

    called := false
    oldFn := handlers.GenerateDisplayNameAsync
    handlers.GenerateDisplayNameAsync = func(_, _, _ string, _ handlers.SessionContext) {
        called = true
    }
    defer func() { handlers.GenerateDisplayNameAsync = oldFn }()

    msgs := []types.Message{
        {ID: "msg-1", Role: "assistant", Content: "I'll help you debug"},
    }

    triggerDisplayNameGenerationIfNeeded("test-project", "test-session", msgs)

    if called {
        t.Error("Expected GenerateDisplayNameAsync to NOT be called when there are no user messages")
    }
}
```

**Standard violated:** specs/standards/backend/conventions.spec.md — tests must actually exercise the code path they claim to cover.

---

### Minor

**2. `getDisplayName` silently discards `found` — `agui_proxy_test.go:247`**

```go
dn, _, _ := unstructured.NestedString(item.Object, "spec", "displayName")
```

If the `spec.displayName` path doesn't exist (e.g., the test session was created with a malformed spec), this returns `""` and the test passes silently for the wrong reason. The `found` bool should be checked in at least the tests that assert a non-empty displayName:

```go
dn, found, err := unstructured.NestedString(item.Object, "spec", "displayName")
if err != nil {
    t.Fatalf("NestedString error: %v", err)
}
_ = found  // acceptable to ignore for empty-check tests
```

---

**3. `var GenerateDisplayNameAsync` loses function name in goroutine stack traces — `display_name.go:~84`**

The conversion from a named function to a `var` means goroutine dumps will show `handlers.GenerateDisplayNameAsync.func1.1` instead of `handlers.GenerateDisplayNameAsync`. This is a known Go debugging trade-off when using function variables. Worth a brief note in the comment block so future readers understand the intent:

```go
// var allows test injection; see agui_proxy_test.go for usage.
var GenerateDisplayNameAsync = func(...) {
```

---

## Positive Highlights

- The PR description is excellent — the two-guard analysis (`ShouldGenerateDisplayName` + first-write-wins) is exactly the kind of reasoning that makes a change trustworthy. The worst-case cost estimate is a nice touch.
- `TestTriggerDisplayName_InitialPromptNotSkipped` is the right test for the core regression — it positively verifies the function is called, not just that the side effect happened.
- `TestTriggerDisplayName_SkipsWhenDynamicClientNil` is a solid defensive test for the nil guard.
- `setupDisplayNameTest` / `cleanup` pattern with deferred restore keeps global state tidy.
- The `var` injection approach is consistent with the existing `GetAgenticSessionV1Alpha1Resource` pattern already in the codebase — no new pattern introduced.

---

## Recommendations

1. Fix `TestTriggerDisplayName_SkipsWhenNoUserMessage` to use the mock, so the test actually verifies the skip. (Major — should fix before merge.)
2. Address the `found` discard in `getDisplayName` for cleaner test hygiene. (Minor.)
3. Add a one-line comment to the `var GenerateDisplayNameAsync` declaration noting the test-injection intent. (Minor.)

---
*Review by Amber — Ambient Code automated review agent*

PR #1566, opened 2026-05-12
feat(cli,sdk,manifests): credential name resolution + platform ingress NetworkPolicy

markturansky:
## Summary

- **Credential name resolution**: All `acpctl credential` subcommands (`get`, `update`, `delete`, `token`) now accept credential names in addition to IDs, with automatic name→ID resolution via `FindByName`
- **SDK FindByName**: Added `FindByName()` to `CredentialAPI` in the Go SDK using TSL search (`name = '<name>'`)
- **Apply uses FindByName**: `acpctl apply` credential upsert now uses `FindByName` instead of `Get` (which only accepts IDs)
- **Delete name resolution**: Top-level `acpctl delete credential <name>` resolves name→ID before deletion
- **Describe/Get fallback**: Top-level `acpctl describe credential` and `acpctl get credential` fall back to `FindByName` when `Get` by ID fails
- **Platform ingress NetworkPolicy**: Adds ingress rule allowing traffic from pods with `ambient-code.io/managed: "true"` label across namespaces
- **project_id compatibility**: SDK `Create` retains `project_id` injection for backward compat with deployed server (TODO to remove after migration `202505120001` is deployed)
- **Fix credential token 404**: Runner now uses the correct global credential path (`/api/ambient/v1/credentials/{id}/token`) matching the API server route, instead of the non-existent project-scoped path that returned 404
- **Project-scoped credential binding**: Control plane `resolveCredentialIDs` now queries role bindings filtered by `scope=credential` and `project_id`, so runners only receive credentials explicitly bound to their project — not every credential in the system
- **Demo script**: Added `demo-credentials.sh` replacing the stale `demo-github.sh` with a spec-compliant credential lifecycle demo (create 3 credentials, bind to project, verify in session)

## Tested against live cluster

 < /dev/null |  Operation | Command | Result |
|---|---|---|
| List | `acpctl credential list` | ✅ |
| Create | `acpctl credential create --name ... --provider ...` | ✅ |
| Get by name | `acpctl credential get <name>` | ✅ |
| Get by ID | `acpctl credential get <id>` | ✅ |
| Describe by name | `acpctl describe credential <name>` | ✅ |
| Update by name | `acpctl credential update <name> --description ...` | ✅ |
| Token by name | `acpctl credential token <name>` | ✅ |
| Delete by name | `acpctl credential delete <name> --confirm` | ✅ |
| Bind | `acpctl credential bind <name> --project <project>` | ✅ |
| Apply YAML | `acpctl apply -f -` (Credential) | ✅ |

## Test plan

- [x] All credential CRUD operations tested against ROSA cluster via port-forward
- [x] Name resolution works for all subcommands (get, update, delete, token)
- [x] ID-based lookups still work (backward compatible)
- [x] `credential bind` creates RoleBinding with `scope=credential`
- [x] `go build` succeeds with no errors
- [ ] Verify NetworkPolicy allows platform-managed pod traffic in staging
- [ ] Verify runner credential token fetch succeeds (no more 404) on hcmai
- [ ] Verify unbound credentials are NOT injected into runner pods

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -138,6 +138,7 @@ scripts/feedback-loop/.last-run
 **/minio-credentials-secret.yaml
 **/postgresql-credentials-secret.yaml
 **/unleash-credentials-secret.yaml
+.secrets/
 
 # One-off analysis and experiments
 repomix-analysis/
--- a/components/ambient-cli/cmd/acpctl/apply/cmd.go
+++ b/components/ambient-cli/cmd/acpctl/apply/cmd.go
@@ -243,7 +243,7 @@ func applyProject(ctx context.Context, client *sdkclient.Client, doc resource) (
 }
 
 func applyCredential(ctx context.Context, client *sdkclient.Client, doc resource) (applyResult, error) {
-	existing, err := client.Credentials().Get(ctx, doc.Name)
+	existing, err := client.Credentials().FindByName(ctx, doc.Name)
 	if err != nil {
 		token := os.ExpandEnv(doc.Token)
 		builder := sdktypes.NewCredentialBuilder().
@@ -271,7 +271,7 @@ func applyCredential(ctx context.Context, client *sdkclient.Client, doc resource
 		if buildErr != nil {
 			return applyResult{}, buildErr
 		}
-		if _, createErr := client.Credentials().Create(ctx, cred); createErr != nil {
+		if _, createErr := client.Credentials().CreateCompat(ctx, cred); createErr != nil {
 			return applyResult{}, createErr
 		}
 		return applyResult{Kind: "Credential", Name: doc.Name, Status: "created"}, nil
--- a/components/ambient-cli/cmd/acpctl/credential/cmd.go
+++ b/components/ambient-cli/cmd/acpctl/credential/cmd.go
@@ -9,6 +9,7 @@ import (
 	"github.com/ambient-code/platform/components/ambient-cli/pkg/config"
 	"github.com/ambient-code/platform/components/ambient-cli/pkg/connection"
 	"github.com/ambient-code/platform/components/ambient-cli/pkg/output"
+	sdkclient "github.com/ambient-code/platform/components/ambient-sdk/go-sdk/client"
 	sdktypes "github.com/ambient-code/platform/components/ambient-sdk/go-sdk/types"
 	"github.com/spf13/cobra"
 )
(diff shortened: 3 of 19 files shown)


markturansky (commented):
## Amber Review — LGTM ✓

**PR:** feat(cli,sdk,manifests): credential name resolution + platform ingress NetworkPolicy
**Verdict:** LGTM

### Summary
Multi-component PR delivering three independent improvements: (1) name-based credential lookup across CLI commands, (2) NetworkPolicy tightening for platform ingress, (3) pinned CodeRabbit binary download replacing curl-pipe-sh. All changes are coherent and individually sound.

### Key Findings

#### Security Improvements (notable positives)
- **NetworkPolicy**: Old base policy had `- {}` (allow-all ingress). New policy replaces it with four specific `from` selectors (OpenShift ingress namespace, same-namespace pods, runner pods, managed pods). This is a meaningful security tightening.
- **Dockerfile**: Replaced `curl ... | sh` with a pinned `0.5.0` binary download. Eliminates install-time supply chain risk.
- **Control plane `resolveCredentialIDs`**: Now filters by `project_id` via role bindings rather than listing all credentials globally. Fixes a scope leak where credentials from other projects could be selected.

#### Checks
- ✓ B1: No `panic()`
- ✓ B6: No credential tokens in logs (auth.py change removes one log path, the remaining path logs `len(token)` elsewhere)
- ✓ B7: `%w` used in new error wraps

### Minor Notes (non-blocking)
1. **`FindByName` search query** (`credential_extensions.go`): uses `fmt.Sprintf("name = '%s'", name)` — if a credential name contains a single quote the search may break or be misrouted. The backend's search parser determines actual risk; for a CLI this is operator-controlled input, but worth hardening if the backend supports parameterized queries.
2. **`CreateCompat` shim** has no issue/ticket tracking for removal. The comment says "Remove once all environments run the updated server" — consider adding a `// TODO(#XXXX)` reference so it doesn't drift.
3. **`api-server-security-patch.yaml`** sets `runAsNonRoot: false` for the kind overlay — expected for local dev, but the patch file has no comment explaining why. Worth a one-liner so future maintainers don't assume it was an oversight.
4. **`resolveCredentialID` in `credential/cmd.go`** does two round-trips (ID then name). The pattern in `delete/cmd.go` tries name first. Minor inconsistency — functionally equivalent but harder to maintain.

### Verdict
All changes net-positive. Security posture improved across three axes (NetworkPolicy, Dockerfile, credential scoping). Name resolution is clean and well-encapsulated via `resolveCredentialID`. Minor notes above are non-blocking.

*— Amber (code review agent)*

PR #1569, opened 2026-05-12
feat(security): SSO/JWT authentication migration (Phase 1)

jsell-rh:
## Summary

SSO authentication migration — replaces OpenShift OAuth proxy with direct OIDC authentication via Keycloak. Includes a validated end-to-end native SSO deployment path (api-server + control plane) that bypasses the legacy backend and operator entirely.

### What's implemented

- **Frontend BFF OIDC**: Next.js acts as confidential OIDC client. Browser gets httpOnly session cookie, never sees a raw JWT. Authorization Code Flow with PKCE. Transparent token refresh (5-min access tokens, 30-min sessions).
- **Backend JWT validation**: JWKS-based validation via `lestrrat-go/jwx/v2`. Validates signature, expiration, issuer, and audience.
- **K8s impersonation**: Backend SA + `Impersonate-User`/`Impersonate-Group` headers preserve all existing RBAC enforcement without cluster OIDC federation. Prefers `preferred_username` over `email` for impersonation (matches Keycloak identity brokering claim order). Defaults to `system:authenticated` group when JWT has no groups.
- **Dual-path auth**: JWT validation first, K8s TokenReview fallback for API keys (SA tokens). Both paths use impersonation.
- **SSO_ENABLED env var**: SSO can be enabled via env var without depending on Unleash feature flag.
- **API server JWK_CERT_URL**: Production environment no longer hardcodes the JWKS URL. Configurable via `JWK_CERT_URL` env var or `--jwk-cert-url` CLI flag, enabling non-RH-SSO OIDC providers (e.g., Keycloak).
- **Session expired UX**: Global 401 detection via React Query cache, blocking dialog with login redirect, no retry storms.
- **E2E test auth**: `client_credentials` grant from Keycloak with K8s SA fallback.
- **Local Keycloak**: Kind cluster includes Keycloak with pre-configured realm, dev users, and protocol mappers. `make kind-sso-toggle` switches between SSO and legacy mode.
- **Native SSO deployment workflow**: Complete workflow for deploying with Keycloak SSO, validated on hcmais cluster. See `workflows/deploy-native-sso.md`.
- **Migration ordering fix**: roleBindings `typedFKMigration` ID changed from `202505130001` to `202603100139` to sort after the table-creation migration it depends on (`202603100138` creates the `role_bindings` table; the old ID sorted before it, breaking fresh databases).

  **Production DB impact verified**: the `ambient-code` staging RDS already has `202505130001` in its migrations table. Shipping `202603100139` means gormigrate treats it as a new migration and runs it. Every statement uses `IF EXISTS`/`IF NOT EXISTS`/`DROP NOT NULL` (all idempotent), so all operations no-op on production. The migrations table will contain both IDs — redundant but harmless. Fresh databases (Kind, new namespaces) will only have `202603100139` in the correct sort position.

### Deployment overlays

Overlays reorganized by cluster under `components/manifests/overlays/hcmais/`:

| Overlay | Purpose |
|---------|---------|
| `hcmais/jsell-sso-poc/` | Native SSO ambient deployment (api-server + control plane, no oauth-proxy) |
| `hcmais/keycloak/` | Keycloak instance deployment |

### Key files

| Area | Files |
|------|-------|
| Spec | `specs/security/sso-authentication.spec.md` |
| Deployment workflow | `workflows/deploy-native-sso.md` |
| API server JWKS config | `components/ambient-api-server/cmd/ambient-api-server/environments/e_production.go` |
| Backend SSO | `components/backend/handlers/sso.go`, `server/server.go` |
| Frontend OIDC | `components/frontend/src/lib/oidc.ts`, `session.ts`, `auth.ts` |
| Overlay (SSO PoC) | `components/manifests/overlays/hcmais/jsell-sso-poc/` |
| Overlay (Keycloak) | `components/manifests/overlays/hcmais/keycloak/` |
| RBAC | `base/rbac/backend-clusterrole.yaml`, `base/rbac/control-plane-*` |

### Default behavior

- `make kind-up` deploys with **legacy auth** (SA token, no Keycloak redirect)
- `make kind-sso-toggle` enables Keycloak OIDC for both frontend and backend
- Existing deployments using oauth-proxy are unaffected — SSO code paths only activate with `SSO_ENABLED=true` or the `sso-authentication` Unleash flag
- API server falls back to `sso.redhat.com` JWKS if `JWK_CERT_URL` is not set

### Native SSO deployment (validated)

The end-to-end path without legacy components was validated on the hcmais cluster:

```
acpctl → Keycloak auth → API Server (REST) → DB
                                ↓ gRPC event
                          Control Plane → Runner Pod → Claude response
                                ↑ gRPC messages
                          Runner → API Server → acpctl
```

Prerequisites for a new environment:
1. Keycloak with 3 clients: `ambient-frontend`, `ambient-control-plane` (service account), `ambient-cli` (public)
2. Identity brokering with RH SSO (optional, for RH user federation)
3. Secrets: `sso-credentials`, `control-plane-oidc`, `ambient-vertex`
4. See `workflows/deploy-native-sso.md` for the complete checklist

## Test plan

- [x] Frontend login via Keycloak → session cookie → JWT forwarded to backend
- [x] Backend validates JWT, impersonates user, RBAC enforced
- [x] API key fallback via TokenReview still works
- [x] Token refresh works silently
- [x] Session expired dialog appears on refresh token expiry
- [x] Logout destroys session + Keycloak single sign-out
- [x] `make kind-sso-toggle` switches between SSO and legacy mode
- [x] Legacy mode (SA token auth) works when SSO is off
- [x] E2E token extraction via Keycloak `client_credentials`
- [x] Native SSO: acpctl login via Keycloak → create project → create session → runner pod → Claude response
- [x] Native SSO: control plane gRPC watch streams connected to api-server
- [x] Native SSO: runner authenticates via CP token server, pushes/watches session messages
- [x] Migration ordering: fresh Kind database bootstraps without errors
- [x] Migration idempotency: existing databases handle re-run safely

## Summary by CodeRabbit

* **New Features**
  * SSO via OIDC/Keycloak with session-based login/logout, secure cookies, and backend JWT validation
  * Dual-mode auth (legacy & SSO) behind a feature flag for gradual rollout
  * Frontend session-expired dialog, explicit Logged Out page, and SSO-aware logout flow
  * Improved E2E helpers to run tests in both legacy and SSO modes

* **Documentation**
  * Updated local dev, Kind/Keycloak setup, E2E, and SSO migration/deployment guides


jsell-rh:
### CI note: 3 pre-existing flaky test failures

The following 3 tests fail intermittently due to shared mutable state (`BaseKubeConfig`) between Ginkgo tests in the handler suite:

- `Projects Handler > Project Lifecycle Management > CreateProject > Should handle existing project gracefully`
- `Projects Handler > Project Namespace Management > Should create namespace with proper labels`
- `Projects Handler > Error Scenarios > Should handle concurrent project creation`

These tests pass in isolation but fail when run as part of the full suite due to test ordering. The SSO changes do not affect this path — `SSOEnabled()` returns `false` without Unleash (test environment), so `getK8sClientsDefault` takes the legacy code path unchanged.

markturansky:
## Review — feat(security): SSO/JWT authentication migration (Phase 1)

Solid implementation overall. The BFF pattern (Next.js as confidential OIDC client, browser never sees a raw JWT) is the right call, and the dual-path auth (JWT → TokenReview fallback) makes incremental rollout safe. A few things worth a second look before merge:

---

### 🔴 Migration ID — verify final state

The commit history shows three conflicting commits on the `roleBindings` migration ID:
1. ID changed `202505130001 → 202603100139` (to sort after the table-creation migration)
2. Then reverted back to `202505130001`
3. Then that revert was itself reverted — net effect: `202603100139` lands

The PR description says _"Idempotent — safe for existing databases where the old ID already ran"_ but does not explain the revert-of-revert. If existing prod databases already have `202505130001` in the migrations table, and this PR ships `202603100139`, the migration **will run again** on those databases. Confirm this is intentional and that the migration body is truly idempotent.

---

### 🟡 `e2e-login` route — production guard needs hardening

`/api/auth/sso/e2e-login` accepts a raw token and injects a session cookie, guarded only by `E2E_TEST_HELPERS` env var. If that var is ever set on a production pod (misconfiguration, copy-paste from a test overlay), it becomes an auth bypass with no other checks. Consider:
- Adding an explicit `process.env.NODE_ENV !== 'production'` guard _in addition to_ `E2E_TEST_HELPERS`, or
- Documenting clearly in the overlay that `E2E_TEST_HELPERS=true` is forbidden in non-test namespaces and adding a CI lint rule to catch it.

---

### 🟡 `preferred_username` vs `email` for impersonation — verify consistency

Two commits pull in opposite directions:
- Main impl: prefers `preferred_username` for `Impersonate-User`
- Fix commit `77548d6`: `getUserSubjectFromContext` should use `userEmail` when creating RoleBindings to match the impersonation identity

The SSAR cache key fix (`ssar_cache.go`) that patches the cross-user cache leak is critical and correct — just make sure the field used for the cache key (`userEmail` or `preferred_username`) matches exactly what's in the `Impersonate-User` header across all code paths. A mismatch here would re-introduce the cache leak.

---

### 🟡 JWK cert URL priority

`JWK_CERT_URL env var > --jwk-cert-url flag > default` — env var overriding an explicit CLI flag is non-standard and surprising. Operators passing `--jwk-cert-url` at deploy time would have the env var silently win. Conventional priority is flag > env var > default.

---

### ✅ Strengths worth calling out

- SSAR cache key fix (cross-user authorization cache leak) — critical bug, correctly fixed
- OIDC discovery TTL (5 min) prevents stale endpoints after Keycloak restarts
- Keycloak proxy through frontend (`/sso/*`) cleanly eliminates the split-URL issuer mismatch — good final design decision
- E2E runs both legacy and SSO modes in CI — strong coverage
- `SSO_ENABLED` env var bypass of Unleash is the right escape hatch for native deployment

---

_Review posted by lead agent (markturansky/hcmai-01)_

markturansky:
## CI Failure — `plugins/credentials` integration tests

Following up on my earlier review: the current CI run has a **failing check** that appears to confirm the migration ordering concern I raised.

**Job:** `API Server Integration Tests` — [view run](https://github.com/ambient-code/platform/actions/runs/26239570996/job/77222019018)

**Failing tests** (all in `plugins/credentials` package):
- `TestCredentialGet`, `TestCredentialPost`, `TestCredentialPatch`, `TestCredentialPaging`, `TestCredentialListSearch`, `TestCredentialListTokenOmitted`, `TestCredentialToken`, `TestCredentialDelete`

**Error** (from logs):
```
integration_test.go:65:
    Error posting object: 404 Not Found
    body: {"reason":"The requested resource '/api/ambient/v1/projects/test-project/credentials' doesn't exist"}
```

This is an **API-level 404**, not an HTTP routing miss — the route exists, but the prerequisite project (`test-project`) doesn't exist when the credentials tests run. The server starts cleanly (`TestServerStarts` passes), so this is a **test setup / migration ordering issue**.

**Likely cause:** The `roleBindings` migration ID change (`202505130001 → 202603100139`) means the migration now runs after all `202505...` migrations. If `test-project` creation depends on the roleBindings migration having run, and that migration is now delayed in the ordering, the credentials plugin tests fail because the project prerequisite is never satisfied.

The `fix(api-server): add placeholder integration test to fix CI` commit (most recent) fixed the `TestMain`-with-no-tests issue but didn't address this underlying failure in the credentials package.

**Suggested fix:** Either confirm the migration ordering is correct for a fresh test database, or add an explicit `t.Skip` / pre-condition check in the credentials integration tests that creates/verifies the test project before running credential operations.

_Lead agent — hcmai-01_

jsell-rh:
## Review response

Thanks for the thorough review @mturansk. All points addressed in `e89e44ae`:

### 🔴 Migration ID
**Verified against production DB.** The `ambient-code` staging RDS has `202505130001` already in the migrations table. Shipping `202603100139` means gormigrate runs it as a "new" migration, but every statement is idempotent (`IF EXISTS`/`IF NOT EXISTS`/`DROP NOT NULL`). All operations no-op. The migrations table will have both IDs — redundant but harmless. Updated PR description with this evidence.

### 🟡 e2e-login hardening
Added `AMBIENT_ENV === "production"` guard in addition to `E2E_TEST_HELPERS`. Both conditions must be met for the route to be disabled — even if `E2E_TEST_HELPERS` leaks into a production overlay, `AMBIENT_ENV=production` blocks it.

### 🟡 preferred_username vs email consistency
Aligned `getUserSubjectFromContext` claim priority to match `buildImpersonatingClients`: `preferred_username > email > sub`. The SSAR cache key uses `claims.Sub` (set at `server.go:232`), which is a stable per-user identifier regardless of which claim is used for impersonation — no cache leak.

### 🟡 JWK cert URL priority
Fixed to conventional ordering: CLI flag > env var > default.

### CI failures
Both were pre-existing:
- **API Server Integration Tests**: `TestMain` with zero test functions (added placeholder, then also fixed the credential route mismatch — routes now registered at both `/credentials/` and `/projects/{id}/credentials/` to match OpenAPI spec)
- **E2E `scheduled-sessions`**: flaky OOTB workflow test, passes on re-run, unrelated to our changes

markturansky (commented):
# Amber Code Review — feat(security): SSO/JWT authentication migration (Phase 1)

## Summary

Substantial, well-structured SSO migration. The core auth path — JWKS-based JWT validation, K8s impersonation, dual-path (JWT + TokenReview) fallback, and encrypted httpOnly session cookie — is architecturally sound and the approach is correct. One critical security flaw in the OIDC callback (open redirect) must be fixed before merge, along with a few major issues around the Keycloak proxy surface and committed credential placeholders.

---

## Findings

### Critical

**Open redirect in OIDC callback** — `components/frontend/src/app/api/auth/sso/callback/route.ts`

```typescript
const returnTo = cookieStore.get("oidc_return_to")?.value || "/";
// ...
return NextResponse.redirect(new URL(returnTo, origin));
```

`new URL(absoluteUrl, base)` ignores `base` when the first argument is already absolute. An attacker can craft `/api/auth/sso/login?returnTo=https://evil.com`; the server stores it in the httpOnly `oidc_return_to` cookie; after successful OIDC login the user is silently redirected to `evil.com`. Classic phishing vector.

**Fix:** validate `returnTo` is a relative path before storing/using it:
```typescript
function safeReturnTo(value: string | undefined): string {
  if (!value) return "/";
  try {
    const url = new URL(value, "http://localhost");
    // Reject if the input was already absolute (has its own host)
    if (new URL(value, "http://localhost").origin !== "http://localhost") return "/";
    return url.pathname + url.search;
  } catch { return "/"; }
}
```

Apply in both `login/route.ts` (before storing to cookie) and `callback/route.ts` (before using from cookie).

---

### Major

**1. CHANGE_ME placeholders committed to cluster overlay** — `components/manifests/overlays/hcmais/jsell-sso-poc/`

`sso-credentials.yaml` has `SSO_CLIENT_SECRET: CHANGE_ME`, `SESSION_SECRET: CHANGE_ME`. `keycloak-admin-secret.yaml` has `password: CHANGE_ME`. `keycloak-db.yaml` has `POSTGRES_PASSWORD: CHANGE_ME`. `keycloak-realm.json` has `"clientSecret": "CHANGE_ME"` for both the frontend client and the RH SSO identity provider.

These are in a cluster-specific overlay intended for production (`hcmais`). If applied without rotation the session cookie encryption key is trivially known. Should either:
- Add a `# REQUIRED: rotate before applying` comment block and a pre-flight check in the deployment workflow
- Or use ExternalSecret / SealedSecret so real values are never committed

The Kind overlay correctly uses `dev-secret-do-not-use-in-prod` labels and is isolated to dev. The hcmais overlay needs the same level of care.

**2. Unrestricted Keycloak proxy** — `components/frontend/src/app/sso/[...path]/route.ts`

This route proxies all GET and POST requests to `SSO_ORIGIN` (Keycloak) with no path allowlisting. It exposes every Keycloak HTTP endpoint to the public internet via the frontend, including the admin REST API (`/auth/admin/...`), token introspection, user endpoints, etc. Keycloak has its own auth on admin routes, but this needlessly widens the attack surface and makes ingress WAF rules harder.

Suggested: restrict to paths the OIDC flow actually needs:
```typescript
const ALLOWED_PATHS = /^(\/realms\/[^/]+\/(protocol\/openid-connect|\.well-known))/;
if (!ALLOWED_PATHS.test(`/${path}`)) {
  return NextResponse.json({ error: "Not found" }, { status: 404 });
}
```

**3. E2E auth bypass reachable with SSO enabled on Kind** — `components/manifests/overlays/kind/frontend-test-patch.yaml`

`E2E_TEST_HELPERS=true` is set unconditionally in the Kind test patch. The e2e-login route's second guard (`AMBIENT_ENV === "production"`) is only meaningful if production overlays explicitly set `AMBIENT_ENV=production` on the frontend deployment. The Kind overlay doesn't set it, so after `make kind-sso-toggle` (SSO on), the auth bypass endpoint is live at `/api/auth/sso/e2e-login` for anyone who can reach the Kind cluster and knows the route exists.

Dev-only cluster, so production impact is nil. But a developer running Kind SSO mode with external access (e.g., port-forwarded to a shared network) is exposed. Consider removing `E2E_TEST_HELPERS` from the standard Kind patch and requiring it to be set via a separate `make kind-e2e-mode` toggle, or add `AMBIENT_ENV=development` to the route guard so the check is explicit both ways.

---

### Minor

**4. Module-level OIDC config cache** — `components/frontend/src/lib/oidc.ts:5-9`

`cachedConfig` and `cachedAt` are module-level variables. In serverless / Next.js edge deployments, module state is not guaranteed to persist across invocations, which could cause repeated OIDC discovery on every request. Consider using `lru-cache` or an explicit cache tied to the Next.js request lifecycle, or document that this is intentional for long-lived Node.js processes only.

**5. Silent session destroy on missing refresh token** — `components/frontend/src/lib/session.ts:60-63`

When the access token is expired and there's no refresh token, `session.destroy()` is called silently. No log entry makes it hard to debug "why did my session disappear?" Add: `console.warn("SSO: session expired with no refresh token, destroying")`.

**6. Kind Keycloak SecurityContext** — `components/manifests/overlays/kind/keycloak-deployment.yaml:16`

`runAsNonRoot: false` for Kind Keycloak and Kind api-server (`kind/api-server-security-patch.yaml`). CLAUDE.md mandates `runAsNonRoot` on all containers. Dev-only, so no prod risk, but worth noting the deliberate divergence with an inline comment explaining why Keycloak requires it (`# Keycloak upstream image requires root; acceptable in Kind dev only`).

**7. Empty smoke test** — `components/ambient-api-server/test/integration/integration_test.go:258-261`

`TestServerStarts` has no assertions — the comment explains the intent, but this pattern is fragile: a future refactor that skips `TestMain` would leave a silently-passing test. Consider `require.NotNil(t, helper.AppConfig())` or equivalent to make the assertion explicit.

---

## Positive Highlights

- **JWT validator** is well-built: JWKS cache with 5-min min-refresh, proper issuer/audience/expiration validation, manual issuer check to support alt-issuers (dev vs. public URL). The `validator_test.go` is thorough — expired, tampered signature, wrong issuer/audience, malformed, empty, minimal claims — all covered.
- **K8s impersonation** approach is the right call: no cluster OIDC federation required, all existing RBAC bindings continue to work, SA tokens (API keys) accepted via TokenReview fallback. Identity claim resolution order (`preferred_username > email > sub`) is consistent across `buildImpersonatingClients`, `setIdentityFromClaims`, and `getUserSubjectFromContext`.
- **SSAR cache key** improved from raw bearer token to OIDC sub — reduces cache churn on token rotation and is more correct semantically.
- **`InitJWTValidator` non-fatal design**: server starts normally without SSO configured; the feature flag is also off in that state. Clean.
- **Session cookie** correctly configured: httpOnly, SameSite=lax, encrypted via iron-session.
- **PKCE+state** implemented correctly for Authorization Code Flow.
- **Migration ID change** well-reasoned and idempotency analysis is rigorous.
- **`forwardedIdentityMiddleware`** falls through to legacy header-based path on JWT failure, preserving backward compatibility.

---

## Recommendations (priority order)

1. **Fix open redirect** in `callback/route.ts` and `login/route.ts` — validate `returnTo` is relative before storing/using. (~10 lines)
2. **Add path allowlist** to `sso/[...path]/route.ts` — block everything except `/realms/*/protocol/openid-connect/*` and `/.well-known/*`.
3. **Address CHANGE_ME placeholders** in `hcmais/jsell-sso-poc/` — add deployment workflow gate or switch to ExternalSecret/SealedSecret.
4. **Document / isolate E2E_TEST_HELPERS** — avoid having it enabled by default on the Kind overlay when SSO is toggled on.
5. Items 4–7 are quality improvements that can be addressed in follow-up tickets.

---

*Review by Amber · standards refs: `specs/standards/security/security.spec.md`, `skills/amber-review/SKILL.md`, `CLAUDE.md`*

jsell-rh:
## Response to Amber code review (#4339633631)

All findings addressed in `1d194b00`:

### Critical — Open redirect
Fixed. Added `safeReturnTo()` validation in both `login/route.ts` and `callback/route.ts`. Absolute URLs are rejected and replaced with `/`. Validates by parsing with a localhost base and checking the origin matches.

### Major 1 — CHANGE_ME placeholders
Added warning comment block pointing to `workflows/deploy-native-sso.md` Step 3a.

### Major 2 — Keycloak proxy allowlist
Added `ALLOWED_PATH_PREFIX` regex restricting to `realms/*/protocol/openid-connect`, `.well-known`, `login-actions`, and `resources`. Everything else returns 404.

### Major 3 — E2E auth bypass
Already addressed in prior commit — added `AMBIENT_ENV === "production"` guard. The Kind concern (dev cluster exposure) is valid but low-risk; noted for follow-up.

### Minor items
- **OIDC config cache**: Intentional for long-lived Node.js processes. Serverless edge is not a current target.
- **Silent session destroy**: Added `console.warn` before `session.destroy()`.
- **Kind SecurityContext**: Added inline comment explaining the deliberate divergence.
- **Empty smoke test**: Added real assertions — verifies `testHelper` initialized and `RestURL` returns non-empty.

markturansky (commented):
## Re-review: SSO/JWT Authentication Migration (Phase 1)

> Holistic re-review after fixup commit `1d194b00`. Previous review posted 2026-05-21.

---

### Fixes verified ✅

All six items from the initial review were addressed in the fixup commit. Each one is correctly implemented:

**Open redirect in callback/login routes** — `safeReturnTo()` now resolves all inputs against `http://localhost` and rejects anything whose `origin` differs from that base. Tested mentally against the classic attack vectors: `https://evil.com` → origin mismatch → `/`; `//evil.com/path` → resolves to `http://evil.com/path` → origin mismatch → `/`; `javascript:alert(1)` → URL parse throws → catch → `/`; `/projects/foo` → same origin → `/projects/foo`. Fix is correct.

**Keycloak proxy path allowlist** — regex `/^realms\/[^/]+\/(protocol\/openid-connect|\.well-known|login-actions|resources)\b/` blocks everything except the four legitimate Keycloak endpoint families. Correct.

**E2E auth bypass double guard** — now requires *both* `E2E_TEST_HELPERS === "true"` *and* `AMBIENT_ENV !== "production"` before the bypass activates. Returns 404 otherwise. Correct.

**CHANGE_ME comment on sso-credentials.yaml** — header comment added, clearly marking placeholder values and linking to the deploy workflow. Correct.

**runAsNonRoot in Kind overlay** — inline comment explains why root is acceptable for the Keycloak upstream image in a dev-only context. Correct.

**Integration test guard** — `TestServerStarts` now fails fast with a clear message if `TestMain` didn't initialize the helper. Correct.

---

### Remaining gaps ⚠️

**1. CHANGE_ME comment missing from three sibling overlay files (minor → should fix)**

`sso-credentials.yaml` got the `# REQUIRED: Replace CHANGE_ME values before deploying` header, but the other three files in the same overlay directory did not:

- `components/manifests/overlays/hcmais/jsell-sso-poc/keycloak-admin-secret.yaml`
- `components/manifests/overlays/hcmais/jsell-sso-poc/keycloak-db.yaml`
- `components/manifests/overlays/hcmais/jsell-sso-poc/keycloak-realm.json`

All three contain literal placeholder strings that would break a real deployment if applied as-is. Same fix needed: add the required-replace comment at the top of each file.

**2. `E2E_TEST_HELPERS=true` unconditional in Kind patch (minor → should fix)**

`components/manifests/overlays/kind/frontend-test-patch.yaml` still sets `E2E_TEST_HELPERS=true` unconditionally. The E2E route is now double-guarded in code, but anyone running the Kind overlay gets an always-on test-helper surface. This should either be removed from the base Kind patch and injected only in CI test runs, or at minimum carry a comment explaining the intent. As it stands, any developer who applies the Kind overlay locally activates the bypass endpoint.

---

### New bug: `session.idToken` is never stored 🐛 (must fix before merge)

`logout/route.ts` reads `session.idToken` to build the OIDC `end_session_endpoint` URL:

```typescript
// logout/route.ts
const endSessionUrl = await getEndSessionUrl(
  session.idToken,   // ← always undefined
  redirectUrl,
);
```

`getEndSessionUrl` uses `idToken` as the `id_token_hint` parameter, which tells the identity provider which session to terminate. If it is `undefined`, the function falls back to passing only `client_id`, which means Keycloak may not terminate the upstream SSO session — only the local cookie is destroyed. Users will find themselves re-logged-in silently the next time they visit, which defeats single sign-out.

The root cause is in `callback/route.ts`. `exchangeCode()` returns `idToken`, but the callback only persists three of the four fields:

```typescript
// callback/route.ts (current)
const tokens = await exchangeCode(code, codeVerifier, redirectUri);
session.accessToken  = tokens.accessToken;
session.refreshToken = tokens.refreshToken;
// session.idToken is dropped here   ← bug
session.expiresAt    = tokens.expiresAt;
await session.save();
```

Fix is one line:

```typescript
session.idToken      = tokens.idToken;
```

This must be added before `session.save()` in the callback route.

---

### Style nit (optional)

`safeReturnTo()` is copy-pasted identically into both `login/route.ts` and `callback/route.ts`. It should live in a shared module (e.g. `src/lib/auth-utils.ts`) and be imported from both routes. Not a blocker, but worth cleaning up before this ships to production to avoid the two copies drifting.

---

### Verdict

The fixup commit is a solid response to the initial review — all six critical and major issues are correctly addressed. Two minor gaps remain (CHANGE_ME comments on sibling files, unconditional `E2E_TEST_HELPERS` in Kind patch), and there is one new must-fix bug: **`session.idToken` is never persisted to the iron-session cookie**, which silently breaks OIDC single sign-out.

**Request changes on the `idToken` omission.** The other items can be addressed in the same commit. Happy to re-review once those three lines are in.

jsell-rh:
## Response to re-review

All items addressed in `05bb922e`:

**idToken bug (must-fix)**: Added `session.idToken = tokens.idToken` in the callback route. Single sign-out via `id_token_hint` will now work correctly.

**CHANGE_ME on sibling files**: Added warnings to `keycloak-admin-secret.yaml`, `keycloak-db.yaml`, and `keycloak/kustomization.yaml` (JSON can't have comments, so noted there instead).

**safeReturnTo dedup**: Extracted to `src/lib/auth-utils.ts`, imported from both routes.

**E2E_TEST_HELPERS comment**: Added security note to the Kind overlay explaining the double-guard and advising against shared-network exposure.

markturansky (commented):
## Third Pass — Thorough Line-by-Line Review

> Reviewed after commit `05bb922e`. All three previous fixup commits now absorbed.

---

### All prior issues confirmed resolved ✅

Every finding from reviews 1 and 2 is correctly addressed. I walked each file end-to-end below; I won't rehash what's already fixed.

---

## New findings

### MAJOR — hcmais Keycloak realm missing `groups` scope

**File:** `components/manifests/overlays/hcmais/keycloak/keycloak-realm.json`

The Kind realm correctly defines a realm-level `groups` clientScope with an `oidc-group-membership-mapper`, and all three clients include it in `defaultClientScopes`. The hcmais realm does neither:

```
Kind realm:    ambient-frontend defaultClientScopes: ['openid', 'email', 'profile', 'groups'] ✅
hcmais realm:  ambient-frontend defaultClientScopes: ['openid', 'email', 'profile']         ✗
```

There is no `groups` clientScope definition at the realm level and no per-client protocol mapper for group membership. Tokens issued by the hcmais Keycloak will have no `groups` claim.

In `buildImpersonatingClients` (`handlers/sso.go`):

```go
groups := claims.Groups
if len(groups) == 0 {
    groups = []string{"system:authenticated"}
}
```

Every SSO user on hcmais will be impersonated as `system:authenticated` and nothing else. Any K8s RoleBinding bound to a specific group (e.g., an LDAP-synced group like `ambient-users`) will not be honoured. Users may see unexpected 403s on resources gated by group membership.

**Fix:** Add the `groups` clientScope to the hcmais realm, mirroring the Kind config:

```json
"clientScopes": [
  {
    "name": "groups",
    "protocol": "openid-connect",
    "protocolMappers": [{
      "name": "groups",
      "protocolMapper": "oidc-group-membership-mapper",
      "config": {
        "full.path": "false",
        "access.token.claim": "true",
        "id.token.claim": "true",
        "claim.name": "groups"
      }
    }]
  }
]
```

And add `"groups"` to `ambient-frontend.defaultClientScopes`.

---

### MAJOR — Identity claim precedence is inconsistent across the stack

There are three places that resolve "who is this user"; all three use a different priority order:

| Site | Priority order |
|---|---|
| `buildImpersonatingClients` (sso.go) | `preferred_username` → `email` → `sub` |
| `setIdentityFromClaims` — `userID` (server.go) | `email` → `sub` |
| `setIdentityFromClaims` — `userName` (server.go) | `preferred_username` → `email` |

K8s impersonation uses `preferred_username` as the primary. The context `userID` uses `email` as the primary. For any user that has both claims (the common case), the K8s identity is their `preferred_username` (e.g. `jsell`) but the application `userID` is their `email` (e.g. `jsell@redhat.com`). These two values are used in different places — `userID` for secret key naming, credential lookup; K8s identity for RBAC. If any RBAC RoleBinding is authored with the email address as the subject, it would match K8s impersonation only when `preferred_username` is absent. If it is authored with the username, `userID`-keyed lookups won't match.

The root fix is to make `setIdentityFromClaims` use `preferred_username` first for `userID`, matching the impersonation order:

```go
// Consistent with buildImpersonatingClients
primaryID := claims.PreferredUsername
if primaryID == "" {
    primaryID = claims.Email
}
if primaryID == "" {
    primaryID = claims.Sub
}
c.Set("userID", SanitizeUserID(primaryID))
c.Set("userIDOriginal", primaryID)
```

---

### MINOR — SSO middleware falls through to X-Forwarded-* headers after JWT failure

**File:** `components/backend/server/server.go`, `forwardedIdentityMiddleware`

When SSO is enabled and JWT validation fails (e.g., the request carries an API key that is not a Keycloak JWT), the middleware logs the failure and falls through to the legacy OAuth proxy path:

```go
// JWT validation failed — fall through to header-based extraction
// (API keys will be handled by getK8sClientsDefault via TokenReview)
// [falls through to reading X-Forwarded-User, X-Forwarded-Email, etc.]
```

The comment is correct — API key identity is re-established by `setIdentityFromTokenReview` inside `getK8sClientsDefault`, which overwrites whatever the middleware set. So in practice no handler sees wrong identity. But the middleware shouldn't trust `X-Forwarded-*` headers at all when SSO is the active auth mode — those headers belonged to the OpenShift OAuth proxy era and should be treated as untrusted in the SSO path.

The minimal fix is a guarded return:

```go
if SSOEnabled() {
    // JWT failed AND we're in SSO mode — don't fall through to OAuth proxy headers.
    // Identity will be established by tokenReviewIdentity in getK8sClientsDefault.
    c.Next()
    return
}
```

This is defense-in-depth: it closes the theoretical path where a direct-to-backend caller sets crafted `X-Forwarded-*` headers alongside an API key token.

---

### MINOR — OIDC authorization request only asks for `openid` scope

**File:** `components/frontend/src/lib/oidc.ts`, `buildAuthorizationUrl`

```typescript
scope: "openid",
```

The code later depends on `email` and `preferred_username` claims. These are delivered by Keycloak because `email` and `profile` are in the client's `defaultClientScopes`, but the request doesn't ask for them explicitly. If the Keycloak admin ever changes the default scopes, the claims silently disappear and auth degrades in a hard-to-diagnose way. Should be:

```typescript
scope: "openid email profile",
```

---

### MINOR — `getAccessToken()` doesn't persist refreshed `idToken`

**File:** `components/frontend/src/lib/session.ts`

`refreshOIDCTokens` returns a new `idToken` (Keycloak always issues a new one on refresh). The refresh path saves `accessToken`, `refreshToken`, and `expiresAt` but drops `idToken`:

```typescript
session.accessToken  = tokens.accessToken;
session.refreshToken = tokens.refreshToken;
// session.idToken not updated here
session.expiresAt    = tokens.expiresAt;
```

The old `idToken` from initial login continues to work as `id_token_hint` at logout (Keycloak accepts it), so this isn't broken today. But Keycloak rotates session keys on refresh, and some configurations invalidate old ID tokens. Add `session.idToken = tokens.idToken` to the refresh block for correctness.

---

### MINOR — `ambient-cli` client lacks an audience mapper

**File:** `components/manifests/overlays/hcmais/keycloak/keycloak-realm.json` (and Kind equivalent)

The `ambient-frontend` client has a protocol mapper that adds `ambient-frontend` to the `aud` claim of access tokens. The `ambient-cli` client has no such mapper.

The backend validates JWT audience against `SSO_AUDIENCE` (which is `ambient-frontend`). A user who authenticates via the CLI client will receive a token without `aud: ambient-frontend`, and the backend will reject it with a validation error. If the CLI is intended to reach the backend in Phase 1 (or Phase 2), this needs an audience mapper on the CLI client. If the CLI is out of scope for now, add a comment in the realm JSON noting this.

---

### MINOR — `new URL(SSO_ISSUER_URL)` at module init in Keycloak proxy

**File:** `components/frontend/src/app/sso/[...path]/route.ts`

```typescript
const SSO_ORIGIN = process.env.SSO_ISSUER_URL
  ? new URL(process.env.SSO_ISSUER_URL).origin
  : null;
```

This executes at module load time. If `SSO_ISSUER_URL` is set to a malformed value (missing protocol, trailing space, etc.), `new URL()` throws a `TypeError` that propagates as an unhandled module-load error, crashing the Next.js worker. Wrap it:

```typescript
function parseSSOOrigin(): string | null {
  try {
    return process.env.SSO_ISSUER_URL
      ? new URL(process.env.SSO_ISSUER_URL).origin
      : null;
  } catch {
    console.error('SSO_ISSUER_URL is not a valid URL:', process.env.SSO_ISSUER_URL);
    return null;
  }
}
const SSO_ORIGIN = parseSSOOrigin();
```

---

## Nits

**Keycloak proxy Location rewrite uses `String.replace` (first match only):**

```typescript
value.replace(SSO_ORIGIN, request.nextUrl.origin + "/sso")
```

`String.replace` with a string argument replaces only the first occurrence. Use `replaceAll` or a global regex. Unlikely to matter in practice (Location headers don't repeat the origin), but correct API is `replaceAll`.

**hcmais realm client secret in `keycloak-realm.json` is `CHANGE_ME`:**

If the realm JSON is imported into Keycloak as-is (which the `--import-realm` flag does on startup), the `ambient-frontend` Keycloak client secret will be set to `CHANGE_ME`, separate from the `SSO_CLIENT_SECRET` in the K8s secret. Both must be updated together. The kustomization now has a comment about this; consider adding a line specifically calling out the realm JSON client secret requirement: `"secret": "<must match SSO_CLIENT_SECRET>"`.

---

## Overall assessment

The core implementation is solid. The JWT validator, JWKS cache, K8s impersonation, BFF OIDC session model, and E2E bypass guards are all well-designed. The two major findings (groups scope gap and identity precedence inconsistency) should be addressed before this goes to production — they don't block the PoC deployment but will cause real RBAC failures when users with group-based access controls hit the system.

The five minor items are fixes, not blockers for the PoC. Requesting changes on the groups scope and identity precedence; everything else is advisory.

markturansky:
## Fourth-pass review — amber

Reviewing 5 commits since `05bb922e` (third pass):
- `235bc38` — exclude sso-credentials from kustomize
- `f00fbaf` — idToken in separate cookie to avoid iron-session 4096-byte limit
- `6346433` — fix logout redirect loop, add `/logged-out` landing page
- `5da471a` — address all third-pass findings (identity precedence, groups scope, proxy hardening)
- `f23c8bc` — use Next.js Link on logged-out page

---

### ✅ All third-pass findings resolved

| Finding | Status |
|---------|--------|
| [MAJOR] Identity claim precedence inconsistency (`setIdentityFromClaims` vs `buildImpersonatingClients`) | Fixed — both now `preferred_username > email > sub` (`server.go:188-197`) |
| [MAJOR] hcmais realm missing `groups` clientScope + `oidc-group-membership-mapper` | Fixed — `clientScopes` block added with mapper; both clients have `groups` in `defaultClientScopes` |
| [MINOR] X-Forwarded-* fallthrough when SSO is enabled and JWT fails | Fixed — `server.go:133-137`: explicit early return when JWT validation fails in SSO mode |
| [MINOR] OIDC scopes only requested `openid` | Fixed — `oidc.ts:55`: `scope: "openid email profile"` |
| [MINOR] `getAccessToken()` didn't persist refreshed idToken | Fixed — `session.ts:51-57`: refreshed idToken stored to `oidc_id_token` cookie |
| [MINOR] ambient-cli lacked audience mapper | Fixed — audience mapper added in both Kind and hcmais realm JSON |
| [NIT] Location rewrite used `replace` not `replaceAll` | Fixed — `route.ts:60`: `replaceAll` |
| [NIT] `SSO_ISSUER_URL` parse crash at module init | Fixed — `parseSSOOrigin()` wraps in try/catch, returns `null` → 503 |

---

### ✅ New commits look good

**`f00fbaf` — idToken separate cookie**  
Correct approach for the 4096-byte iron-session limit. The `oidc_id_token` cookie is `httpOnly`, `secure` in production, `sameSite: lax`, `maxAge: 86400`. Logout reads and deletes it before calling Keycloak `end_session_endpoint`. One note: if token refresh fails and `session.destroy()` is called in the error path (`session.ts:62`), the `oidc_id_token` cookie is not cleared — it will persist up to 24 hours. Low risk (idToken alone doesn't grant platform access), but worth noting.

**`6346433` — logout redirect loop fix**  
Root cause correctly identified: logout was redirecting to `/` which auto-re-authenticated via Keycloak session. Now redirects to `/logged-out`. Middleware matcher now excludes `logged-out` (`middleware.ts` config). Clean fix.

**`235bc38` — kustomize exclusion of sso-credentials**  
Same pattern as the CRB fix. Prevents `oc apply -k` from clobbering manually-set secrets with `CHANGE_ME` values. Correct.

**`f23c8bc` — Link component**  
Standard Next.js best practice. Good.

---

### Notes (no action required)

1. **`SSO_ORIGIN` is still eagerly evaluated at module init** — it's now resilient to bad URLs (returns `null` → 503 instead of crashing), but if `SSO_ISSUER_URL` is unset at module load time, the proxy will serve 503 until the module is reloaded. Non-issue in K8s (env vars are set at pod start), but if this ever runs in a hot-reload or serverless context, the lazy init pattern (evaluate inside the handler) would be more robust.

2. **X-Forwarded-* still trusted in SSO mode when no Bearer token is present** — This is the intended API key / SA token path. My original concern was about JWT *failure* fallthrough, which is now fixed. When there's no token at all in SSO mode, X-Forwarded-* → TokenReview is the correct fallback chain.

---

### Verdict: **LGTM** ✅

All major and critical findings across four review passes are resolved. The implementation is sound: PKCE + iron-session BFF, consistent identity claim ordering, groups propagation to Keycloak JWTs, proper X-Forwarded-* isolation in SSO mode, and clean logout flow. The draft status can be lifted when ready.

_Amber — hcmai-01_

PR #1572, opened 2026-05-13
  fix(operator): apply feature flags when scheduled sessions trigger

### LINKED ISSUE #1506: Bug: Intermittent failure to register jira-write MCP tools in active sessions
## 🐛 Bug Description

**Summary:** Intermittent failure to register `jira-write` MCP tools in scheduled sessions.

**Expected Behavior:** When the `jira-write` setting is toggled **On**, all Jira write tools (e.g., `jira_add_comment`, `jira_update_issue`) should be registered and callable by the agent.

**Actual Behavior:** The agent reports that only read operations are available in the `mcp__mcp-atlassian__*` namespace, claiming write tools are not registered as callable functions despite the setting being enabled.

## 🔄 Steps to Reproduce

1. Go to **Workspace Settings** -> **Jira Integration**.
2. Toggle **jira-write** to **On**.
3. Initiate a workflow requiring Jira write access (e.g., run `/triage --comment`).
4. Observe the agent response stating `jira_add_comment` is not a registered tool.

## 🖼️ Screenshots

Image

Image

janisz:
Fixes #1506 - Scheduled sessions were not respecting workspace feature flags like `jira-write`.

  ### Root Cause

  The session-trigger command creates AgenticSession CRs directly from a pre-stored template without re-evaluating workspace feature flags at  trigger time. When regular sessions are created via the backend API, they check the `jira-write` flag and set `JIRA_READ_ONLY_MODE=false` if  enabled. Scheduled sessions skipped this check entirely.

  ### Solution

  Added `applyFeatureFlagOverrides()` function that:
  1. Reads the `feature-flag-overrides` ConfigMap from the workspace namespace
  2. Checks if `jira-write` is set to `"true"`
  3. Sets `JIRA_READ_ONLY_MODE=false` in the session template's environment variables before creating the CR

  The implementation follows the same pattern as `CreateAgenticSession` in the backend: read ConfigMap, check flag value, inject environment  variable. Errors are logged but non-fatal to ensure degraded operation.

## Summary by CodeRabbit

* **New Features**
  * Workspace-level feature-flag overrides can now toggle Jira write behavior, updating session environment variables accordingly.

* **Bug Fixes**
  * Reading missing or unreadable workspace overrides no longer blocks session creation; warnings are logged and creation continues.

* **Tests**
  * Added unit tests verifying override application, preservation of existing environment variables, and non-application for non-true flag values.


markturansky:
# Claude Code Review

## Summary

Clean, targeted fix that correctly ports the feature-flag override pattern from `CreateAgenticSession` (backend) to the operator's session-trigger path. The approach is sound: non-fatal ConfigMap reads, typed constants, table-driven tests covering all flag states. No blockers or criticals.

## Findings

### Blocker
None.

### Critical
None.

### Major

**1. `config.K8sClient != nil` guard is misleading after a successful `InitK8sClients()`**
`createNewSession` gates on `config.K8sClient != nil` before calling `applyFeatureFlagOverrides`. But `RunSessionTrigger` calls `config.InitK8sClients()` and fatally exits on error — so if we reach `createNewSession`, `InitK8sClients` already succeeded. A successful `InitK8sClients` should guarantee both `DynamicClient` and `K8sClient` are non-nil. The nil guard therefore implies it's an acceptable degraded state for the typed client to be missing, which contradicts the fatal-exit contract above and means feature flags silently won't apply in a case that should never occur. Either:
  - Remove the guard (trust the fatal-exit invariant), or
  - Make `InitK8sClients` explicitly guarantee the typed client is set and document why the guard exists.

### Minor

**2. RBAC not verified in diff**
`applyFeatureFlagOverrides` adds a `CoreV1().ConfigMaps(namespace).Get()` call from the operator SA. The diff shows no RBAC manifest changes. CodeRabbit notes this was covered in PR #1138 — if that's confirmed, no action needed. But worth a comment in the PR linking to the existing ClusterRole rule so reviewers can verify without archaeology.

**3. Removed explanatory comment on name-budget constraint**
The original `createNewSession` had a comment explaining the 55-char K8s resource name budget derivation (session-N prefix + sanitized name + timestamp). That's a non-obvious invariant — a future reader adding to the name format could easily break it without the comment. Worth reinstating as a single-line note.

### Nit

- `applyFeatureFlagOverrides` always returns `nil` — all error paths log and return `nil`. The `error` return type creates an expectation of recoverable errors that's never fulfilled. Either remove the return type (making non-fatal explicit in the signature) or return actual errors and let the caller decide on severity.
- 15 commits for a focused fix; squash on merge will handle it, but keeping the branch tidy makes `git bisect` easier during development.

## Positive Highlights

- **Correct non-fatal design.** Logging and continuing when the ConfigMap is missing or unreadable is exactly right — don't block session creation for feature-flag degradation.
- **Constants in `types/resources.go`.** `FeatureFlagOverridesConfigMap`, `JiraWriteFlagKey`, `JiraReadOnlyModeEnvVar` are all in the right file alongside peer constants. No magic strings scattered in implementation code.
- **Test coverage is solid.** Six table-driven cases cover: flag enabled, flag disabled, no ConfigMap (NotFound), pre-existing env vars preserved, unrelated flags ignored, non-"true" values rejected. This is the right test surface for the function.
- **No `panic()`, typed clients used correctly, existing env vars are additive not replaced.** Conventions clean.

## Recommendations

1. (Major) Remove the `config.K8sClient != nil` guard or document why a nil typed client after successful init is possible.
2. (Minor) Add a link to the existing RBAC rule in the PR description or a code comment, so the ConfigMap `get` permission is traceable.
3. (Minor) Reinstate the one-line comment about the name-budget constraint in `createNewSession`.

---

*Amber · standards loaded from CLAUDE.md, specs/standards/control-plane/*

markturansky (changes requested):
# Amber Review — PR #1572

## Summary

Correct approach overall — mirrors the backend `CreateAgenticSession` pattern for feature-flag injection into scheduled sessions. However, there is an unresolved structural bug that means `JIRA_READ_ONLY_MODE` is never actually applied at runtime. Both the implementation and its tests have the same mismatch.

## Findings

### Blocker — Wrong nesting level in `applyFeatureFlagOverrides`

`template` is already the AgenticSession **spec** — confirmed by:
- `template["displayName"]` accessed directly in `createNewSession` (not `template["spec"]["displayName"]`)
- `template["initialPrompt"]` accessed directly in `tryReuseLastSession`
- `createNewSession` wraps it as `"spec": template` when building the CR

So `template["spec"]` produces an extra level of nesting: the env var lands at `spec.spec.environmentVariables`. The operator reads `spec.environmentVariables` → `JIRA_READ_ONLY_MODE` is never applied. The feature flag fix is silently a no-op.

**`trigger.go` fix:**
```diff
-	spec, ok := template["spec"].(map[string]interface{})
-	if !ok {
-		spec = map[string]interface{}{}
-		template["spec"] = spec
-	}
-
-	envVars, ok := spec["environmentVariables"].(map[string]interface{})
+	envVars, ok := template["environmentVariables"].(map[string]interface{})
 	if !ok {
 		envVars = map[string]interface{}{}
 	}
 
 	envVars[types.JiraReadOnlyModeEnvVar] = "false"
-	spec["environmentVariables"] = envVars
+	template["environmentVariables"] = envVars
```

### Blocker — Tests validate the wrong contract

The tests build `template = {"spec": map{}}` and assert `template["spec"]["environmentVariables"]`. But the real caller passes a flat spec map with no outer `"spec"` key. The tests exercise the wrong path, so they pass regardless of whether the production bug exists.

**`trigger_test.go` fix:**
```diff
-			template := map[string]interface{}{
-				"spec": map[string]interface{}{},
-			}
+			template := map[string]interface{}{}
 			if tt.existingEnvVars != nil {
-				spec := template["spec"].(map[string]interface{})
-				spec["environmentVariables"] = tt.existingEnvVars
+				template["environmentVariables"] = tt.existingEnvVars
 			}
```

And update all assertions to read from `template["environmentVariables"]` directly (remove the `template["spec"]` intermediary).

## Score

2 blockers. REJECT — the fix is mechanically small but the tests need to be corrected too, otherwise the same class of bug can recur silently.

---
*— Amber (code review agent)*

markturansky:
# Amber Code Review — REJECT

**Reviewing at head SHA `7d0cf5f`** (re-review; prior comment from 2026-05-21 missed these blockers)

---

## Findings

### Blocker 1 — `applyFeatureFlagOverrides` writes env var to the wrong nesting level

In `trigger.go`, `applyFeatureFlagOverrides` does:

```go
spec, ok := template["spec"].(map[string]interface{})
if !ok {
    spec = map[string]interface{}{}
    template["spec"] = spec
}
envVars, ok := spec["environmentVariables"].(map[string]interface{})
...
spec["environmentVariables"] = envVars
```

But `template` **is already the AgenticSession spec**. `createNewSession` builds the CR with:

```go
"spec": template,
```

So writing to `template["spec"]["environmentVariables"]` produces the K8s path `spec.spec.environmentVariables` — a field that does not exist in the CRD schema. `JIRA_READ_ONLY_MODE=false` is never applied to the actual session.

**Fix:**

```diff
-  spec, ok := template["spec"].(map[string]interface{})
-  if !ok {
-      spec = map[string]interface{}{}
-      template["spec"] = spec
-  }
-
-  envVars, ok := spec["environmentVariables"].(map[string]interface{})
+  envVars, ok := template["environmentVariables"].(map[string]interface{})
   if !ok {
       envVars = map[string]interface{}{}
   }

   envVars[types.JiraReadOnlyModeEnvVar] = "false"
-  spec["environmentVariables"] = envVars
+  template["environmentVariables"] = envVars
```

---

### Blocker 2 — Unit tests mirror the production bug and validate the wrong contract

The tests create the fixture as:

```go
template := map[string]interface{}{
    "spec": map[string]interface{}{},
}
```

…and assert against:

```go
spec, ok := template["spec"].(map[string]interface{})
envVars, ok := spec["environmentVariables"].(map[string]interface{})
```

Because the production code and the tests both use the same `template["spec"]` path, all 6 test cases pass green — but they are validating that the env var lands at `spec.spec.environmentVariables`, which is exactly the wrong place. The tests do not exercise the actual runtime contract.

**Fix (test fixture and assertions):**

```diff
-  template := map[string]interface{}{
-      "spec": map[string]interface{}{},
-  }
+  template := map[string]interface{}{}
   if tt.existingEnvVars != nil {
-      spec := template["spec"].(map[string]interface{})
-      spec["environmentVariables"] = tt.existingEnvVars
+      template["environmentVariables"] = tt.existingEnvVars
   }
   ...
-  spec, ok := template["spec"].(map[string]interface{})
-  if !ok {
-      t.Fatal("template[spec] is not a map")
-  }
-  envVars, ok := spec["environmentVariables"].(map[string]interface{})
+  envVars, ok := template["environmentVariables"].(map[string]interface{})
```

---

## Summary

Both blockers share the same root cause: the author assumed `template` wraps a `"spec"` key, but it does not — it **is** the spec. The fix is small (3 lines in production, 5 in tests) but critical: without it, the scheduled-session feature-flag fix does nothing at runtime, and the test suite provides false confidence that it works.

The prior Amber comment (2026-05-21) reviewed the K8s client nil-guard issue and missed these two. That analysis still stands as a Major finding, but these blockers take precedence.

---

*Amber · standards from CLAUDE.md, specs/standards/control-plane/*

janisz:
@coderabbitai full review

markturansky:
🤖 **Amber Analysis** — PR #1572: fix(operator): apply feature flags when scheduled sessions trigger

Solid bug fix. The root cause (scheduled sessions bypassing workspace feature-flag evaluation) is correctly diagnosed, and the follow-up commit catching the nesting bug (`template[\"spec\"][\"environmentVariables\"]` → `template[\"environmentVariables\"]`) is the right fix—that earlier path would have been a silent no-op at runtime.

**Verdict: LGTM** — two minor nits below, neither blocks merge.

---

## Findings

### Blocker
None.

### Critical
None.

### Major
None.

### Minor

**1. `applyFeatureFlagOverrides` has a dead error return** (`trigger.go`)

The function signature is `func applyFeatureFlagOverrides(...) error`, but it never returns a non-nil error—all error paths log a warning and return `nil`. This makes the `if err != nil` branch in `createNewSession` dead code and the `error` return misleading.

Two clean options:
- Drop the return type entirely: `func applyFeatureFlagOverrides(...)` and call it as a statement
- If you want to keep it extensible, add a comment: `// Always returns nil: designed non-fatal; future callers may surface real errors`

Relevant standard: error-handling.spec.md — misleading signatures hide intent; silent-failure anti-pattern applies even at the call site.

**2. Unreachable nil guard in `createNewSession`** (`trigger.go`)

```go
if config.K8sClient != nil {
    // apply feature flags
}
```

`RunSessionTrigger` calls `log.Fatalf` if `config.InitK8sClients()` fails, so `config.K8sClient` is always initialized by the time `createNewSession` is reached. The guard silently skips feature flag application in a scenario that can't happen, which would be harder to debug if the assumption ever changes.

The project standard (CLAUDE.md) is explicit: _"Don't add error handling... for scenarios that can't happen."_ Removing the guard makes the intent clear; if future callers pass an uninitialized client, they should get a panic early rather than a silent skip.

---

## Positive Highlights

- **Nesting bug caught and fixed**: the follow-up commit (`fix(operator): correct env var nesting level`) correctly writes to `template["environmentVariables"]` instead of `template["spec"]["environmentVariables"]`. The updated tests validate the right contract.
- **Constants in `types/resources.go`**: `FeatureFlagOverridesConfigMap`, `JiraWriteFlagKey`, `JiraReadOnlyModeEnvVar` — no stringly-typed code anywhere.
- **6 thorough unit tests** with `fake.Clientset`: covers enabled, disabled, missing ConfigMap, preserve existing vars, unrelated flags, and non-true value. This is the right test surface for an isolated helper.
- **Non-fatal design is correct**: logging a warning and continuing when the ConfigMap is absent is strictly better than blocking session creation. `errors.IsNotFound` handling follows control-plane conventions spec exactly.
- **DRY client init**: reusing `config.InitK8sClients()` + `config.DynamicClient` / `config.K8sClient` instead of creating a new in-cluster config on every trigger invocation is a clear improvement.
- **Comment hygiene**: WHAT comments removed; only the WHY for the non-fatal design decision is retained.

<details>
<summary>Files reviewed</summary>

- `components/operator/internal/trigger/trigger.go` (+45/-24)
- `components/operator/internal/trigger/trigger_test.go` (+114)
- `components/operator/internal/types/resources.go` (+9)

Standards applied: control-plane/conventions.spec.md, error-handling.spec.md, security/security.spec.md

</details>

---

*Confidence: High (95%) — bug fix in a well-understood path, comprehensive tests, follows established patterns.*

markturansky (commented):
## Amber Review — LGTM ✓

**PR:** fix(operator): apply feature flags when scheduled sessions trigger
**Verdict:** LGTM

### Summary
Clean fix for #1506. `applyFeatureFlagOverrides()` correctly reads the workspace `feature-flag-overrides` ConfigMap at trigger time and injects `JIRA_READ_ONLY_MODE=false` when `jira-write: "true"`. The critical nesting bug (writing to `spec.spec.environmentVariables` instead of `spec.environmentVariables`) was caught and fixed before merge. Non-fatal error handling (log+continue) is appropriate for a degraded-operation path.

### Checks Passed
- ✓ B1: No `panic()` in production code
- ✓ B3: No direct type assertions on K8s unstructured objects (template is local map, not `unstructured.Unstructured`)
- ✓ B4: Errors logged, not silently dropped
- ✓ B6: No tokens in logs
- ✓ B7: Uses `%w` where errors propagate
- ✓ Tests: 6 table-driven cases covering enabled, disabled, missing ConfigMap, existing vars preserved, other flags ignored, non-"true" value

### Minor Notes (non-blocking)
1. `applyFeatureFlagOverrides` handles only `jira-write`. If more workspace flags are added, this function grows. A map of `flag → envVar` could scale better, but for one flag YAGNI applies.
2. The function signature returns `error` but always returns `nil` (errors are swallowed with log). The return type is future-proofing — acceptable, but callers should not expect the error path to be reachable today.

*— Amber (code review agent)*

PR #1591, opened 2026-05-15
Delegate urls flag

maknop:
# Summary of Changes

File modified: components/manifests/templates/template-services.yaml:692

  Change:
  - `--scope=user:full`
  + `--openshift-delegate-urls={"/":{"resource":"projects","verb":"list"}}`
  - `--upstream-timeout=5m`

## What This Fixes
The --openshift-delegate-urls flag makes the OAuth proxy validate the user's access token on every request by checking if they have permission to list projects. This ensures:

  1. Expired tokens are caught immediately at the proxy level instead of being forwarded to the backend
  2. Better error messages - users get redirected to re-authenticate instead of seeing "Token expired or invalid"
  3. Consistent behavior with the production overlay (which already has this flag)

Why This Solves the ~2 Minute Issue

  Without this flag:
  - OAuth proxy only validates authentication (logged in?) but not authorization
  - Expired tokens are passed through to the backend
  - Backend discovers the expired token only when SSAR cache expires (30s intervals)
  - Creates the cascading error pattern you were seeing

  With this flag:
  - OAuth proxy validates both authentication AND authorization on every request
  - Expired/invalid tokens are rejected at the proxy layer
  - User is automatically redirected to re-authenticate
  - No cascading errors

## Summary by CodeRabbit

* **New Features**
  * Added OAuth authentication proxy for frontend secure access.
  * Introduced comprehensive CI/CD pipeline configurations for automated component builds.
  * Added deployment templates for operator and application services on Kubernetes.

* **Chores**
  * Upgraded frontend Node.js runtime to version 24.
  * Standardized database secret naming across deployments.
  * Updated database configuration for external RDS connectivity with SSL.


jsell-rh:
@maknop Is this still a desired change after we've identified the root cause for the 2m expiration? 

Further, it looks like the PR might contain more than just the patch described in the PR body?

maknop:
Sorry, this was accidentally opened! I'll close.

PR #1599, opened 2026-05-20
feat(specs): credential sidecar isolation architecture

markturansky:
## Summary

Implements credential sidecar isolation architecture across 7 waves:

- **Wave 1**: CP injects per-credential sidecar containers (GitHub/Jira/K8s/Google) into session pods with `CREDENTIAL_MCP_URLS` env var
- **Wave 2**: Runner connects to credential sidecars via SSE/streamable-http instead of subprocess MCP; skips token injection in sidecar mode
- **Wave 3**: System prompts use MCP tools (`push_files`, `create_pull_request`) instead of `git push`/`gh pr create` when sidecars active
- **Wave 4**: Dockerfiles for all 4 credential sidecars + Makefile build targets
- **Wave 5**: Unit tests for `buildCredentialSidecars` (6 test cases covering all edge cases)
- **Wave 6**: Shared `credential-entrypoint` Go binary for RSA-OAEP token exchange and credential fetching
- **Wave 7**: Cleanup (gitignore built artifacts)

Integration tokens (GitHub, GitLab, Jira, kubeconfig, Google) are isolated in sidecar containers — the runner container has zero integration tokens in sidecar mode. LLM credentials remain in the runner. Full backward compatibility via fallback to subprocess mode.

## Files changed

 < /dev/null |  Component | Files | Purpose |
|---|---|---|
| Control Plane | `config.go`, `kube_reconciler.go`, `main.go` | Sidecar injection, config, wiring |
| Runner | `mcp.py`, `auth.py`, `prompts.py` | SSE transport, token gating, MCP prompts |
| Sidecars | `credential-sidecars/{github,jira,k8s,google}/Dockerfile` | Container images |
| Entrypoint | `credential-sidecars/entrypoint/main.go` | Token exchange wrapper |
| Tests | `kube_reconciler_test.go` | 6 unit tests |
| Build | `Makefile` | `build-credential-sidecars` target |

## Test plan

- [x] CP: `go vet ./...` clean
- [x] CP: `go test ./...` — all pass (6 new + 3 existing)
- [x] Runner: `pytest` — 714 passed, 11 skipped
- [x] GitHub sidecar image builds and starts (`podman build` + `podman run`)
- [ ] Full kind cluster E2E with credential sidecars injected
- [ ] Verify MCP tool connectivity from runner to sidecar
- [ ] Verify token refresh cycle in sidecar entrypoint

## Summary by CodeRabbit

* **New Features**
  * Per-credential, localhost-exposed MCP sidecars for integration credentials; sidecars refresh tokens and can be omitted when a project has no bound credentials.
  * Runner is explicitly credential-free for integrations; git write operations must use MCP tools (cloning moved to an init step).

* **Documentation**
  * Opt-in sidecar deployment, session annotation, local sidecar endpoints, migration notes for tool name changes, and updated security/design invariants (LLM provider credentials exempt).


markturansky:
# Claude Code Review - PR #1599

## Summary

This PR documents a significant architectural shift: moving from in-process credential management to sidecar-based credential isolation. Three spec files are updated to reflect that integration credentials (GitHub, GitLab, Jira, Google, kubeconfig) SHALL NOT be visible to the agent process, while LLM provider credentials (Anthropic/Vertex) remain in the runner container. The specs are well-written, internally consistent, and represent a positive security enhancement.

## Findings

### Blocker

None

### Critical

**1. PR scope mismatch (specs vs. implementation)**  
**File**: N/A (PR metadata)  
**Issue**: The PR title is `feat(specs,cli): credential sidecar isolation specs`, which suggests both spec and CLI changes. However, the diff contains only spec file modifications (`specs/agents/runner.spec.md`, `specs/integrations/mcp-server.spec.md`, `specs/security/security.spec.md`). No CLI implementation changes are present.

The test plan mentions:
```
cd components/ambient-cli && go test ./cmd/acpctl/credential/ ./cmd/acpctl/agent/
```

This implies implementation code exists, but it's not in this PR diff.

**Violated Standard**: CLAUDE.md - "Verify contracts and references"

**Suggested Fix**:
- If this is a **specs-only PR**: Change title to `feat(specs): credential sidecar isolation specs` and remove CLI mention
- If **implementation is missing**: Add the corresponding CLI/backend/operator code changes
- If **implementation is in a separate PR**: Note that in the PR description and ensure the specs-first approach is intentional

### Major

**1. CodeRabbit credential provider not in sidecar table**  
**File**: `specs/security/security.spec.md:170`  
**Issue**: The security spec lists "GitHub, GitLab, Jira, Google, kubeconfig, CodeRabbit" as integration credentials, but the MCP server spec's sidecar table (line 987-992) only includes `github`, `gitlab`, `jira`, `kubeconfig`, and `google`. CodeRabbit is missing.

**Violated Standard**: Cross-cutting convention - "Verify contracts and references. After moving anything, grep scripts, workflows, manifests, and configs"

**Suggested Fix**:
- **Option A**: Add a `coderabbit-mcp` row to the sidecar table in `specs/integrations/mcp-server.spec.md` if CodeRabbit will have a credential sidecar
- **Option B**: Remove "CodeRabbit" from the security spec credential list if it follows a different integration pattern

**2. Removed `acp` MCP server tools without migration path**  
**File**: `specs/agents/runner.spec.md:377-387`  
**Issue**: The diff removes the `acp` in-process MCP server with 9 tools (`acp_list_sessions`, `acp_get_session`, `acp_create_session`, etc.) and replaces them with SSE sidecar architecture. However, there's no documented migration path or backward compatibility strategy for existing sessions or workflows that may depend on these tools.

**Violated Standard**: While not explicitly a standards violation, this is a breaking architectural change that affects existing functionality.

**Suggested Fix**:
- Add a migration section to `specs/agents/runner.spec.md` explaining:
  - How existing sessions using `acp` tools will be handled
  - Whether the tools are simply renamed/moved or deprecated
  - If these 9 tools are now exposed by the `ambient` sidecar (`:8090`), clarify that in the spec

### Minor

**1. GitLab sidecar image marked as TBD**  
**File**: `specs/integrations/mcp-server.spec.md:989`  
**Issue**: The GitLab MCP sidecar row shows `TBD (community or custom)` for the image field, while all other sidecars have specific images or commands.

**Impact**: Low - this is acceptable for a spec document, but signals incomplete design.

**Suggested Fix**: If the image is known, specify it. Otherwise, add a note that this will be determined during implementation.

**2. Incomplete test plan checkboxes**  
**File**: N/A (PR description)  
**Issue**: The test plan has unchecked boxes:
```markdown
- [ ] Verify spec changes are consistent across all three documents
- [ ] `cd components/ambient-cli && go test ./cmd/acpctl/credential/ ./cmd/acpctl/agent/`
```

**Suggested Fix**: Complete the checklist before requesting review or merge. If tests haven't been run, the PR may be premature.

## Positive Highlights

1. **Strong security enhancement**: The credential sidecar isolation model prevents token exfiltration by the agent via `Bash`/`Read` tools. This is a significant security improvement over in-process credential management.

2. **Well-structured scenarios**: The security spec includes clear scenarios (agent cannot read tokens, credential sidecar isolation, git credential proxy) that make the requirements testable and understandable.

3. **Comprehensive sidecar table**: The MCP server spec's provider-to-sidecar mapping table clearly documents image names, ports, and environment variables for each integration.

4. **Explicit LLM credential exemption**: The spec correctly identifies that Anthropic/Vertex credentials must remain in the runner container for inference, with clear rationale.

5. **Updated Design Decisions section**: The security spec's Design Decisions table includes the new sidecar isolation rationale, maintaining the historical decision log.

6. **Good pod layout visualization**: The revised pod layout diagram in `specs/integrations/mcp-server.spec.md` clearly shows the runner container without integration tokens and separate credential sidecars.

## Recommendations

**Priority order:**

1. **CRITICAL**: Clarify PR scope - Either add implementation code or adjust title/description to reflect specs-only change
2. **MAJOR**: Resolve CodeRabbit discrepancy - Add to sidecar table or remove from credential list
3. **MAJOR**: Document migration path for removed `acp` MCP server tools
4. **MINOR**: Specify GitLab sidecar image or add implementation note
5. **MINOR**: Complete test plan checklist before merge

The specs themselves are high-quality and represent sound architectural thinking. Once the scope/consistency issues are resolved, this will be a strong foundation for the sidecar credential isolation implementation.

markturansky:
# Review Update - PR #1599

## Changes Since Initial Review

Excellent progress! The PR author has addressed **2 of 5** findings from the initial review and added valuable clarifications.

## ✅ Issues Addressed

### 1. Migration Path for `acp` Tools (MAJOR → Resolved)

**Original Issue**: Removed `acp` MCP server tools without migration path

**Resolution**: Added comprehensive migration section in `specs/agents/runner.spec.md:384-394`:

```markdown
### Migration: `acp` In-Process MCP Server Removed

The previous `acp` in-process MCP server (9 tools: `acp_list_sessions`,
`acp_get_session`, `acp_create_session`, ...) is replaced by the `ambient` 
SSE sidecar on `:8090`.

The `ambient-mcp` sidecar exposes the same platform tools (sessions, agents,
projects) via the MCP protocol over SSE. Tool names change from `acp_*` prefix
to unprefixed (`list_sessions`, `get_session`, etc.). Existing agent prompts
referencing `acp_*` tool names must be updated.
```

**Verdict**: ✅ Well documented. Clear mapping from old to new tool names.

### 2. GitLab Sidecar Image TBD (MINOR → Improved)

**Original Issue**: GitLab sidecar marked as "TBD (community or custom)"

**Resolution**: Expanded to `"TBD — no official MCP server exists yet; will require a community or custom image"` (line 989)

**Verdict**: ✅ More explicit. Clarifies that the blocker is the absence of an official server, not incomplete design.

## ➕ Valuable Additions

### 1. Git Operations Clarification

**New Section**: `specs/agents/runner.spec.md:351-364` - "Git Operations"

Key points:
- Explicitly states runner has **no git credential helper** and **no GitHub/GitLab tokens**
- Documents MCP tool alternatives: `PushFiles`, `CreatePullRequest`
- Clearly states `git push` and `gh pr create` are **not supported** (would violate isolation)
- References init container for clone operations

**Impact**: Prevents implementation confusion. Makes architectural constraints explicit.

### 2. Enhanced Sidecar Token Refresh Details

**Location**: `specs/integrations/mcp-server.spec.md:997-1002`

Added environment variables for all credential sidecars:
- `AMBIENT_API_URL`
- `AMBIENT_CP_TOKEN_URL`
- `SESSION_ID`

Explains that sidecars use RSA-OAEP token exchange for credential refresh, same mechanism as `ambient-mcp` sidecar.

**Impact**: Complete picture of how sidecars maintain credentials over time.

### 3. Git Operations Security Scenario

**New Scenario**: `specs/security/security.spec.md:195-202` - "Git write operations use MCP tools, not tokens"

Explicitly documents that:
- Agent calls `github-mcp` sidecar's `PushFiles` or `CreatePullRequest` tools
- Runner never has git credential helper or token
- Direct `git push` / `gh pr create` **SHALL fail** (no credentials available)

**Impact**: Makes security boundary testable. Clear pass/fail criteria.

## ⚠️ Outstanding Issues

### 1. PR Scope Mismatch (CRITICAL - Still Open)

**Status**: NOT addressed (requires PR metadata change, not code)

**Issue**: PR title is `feat(specs,cli): credential sidecar isolation specs` but diff contains only spec files, no CLI implementation.

**Action Required**: 
- Change title to `feat(specs): credential sidecar isolation specs`, OR
- Add corresponding CLI implementation to this PR, OR
- Document in PR description that implementation will follow in separate PR

### 2. CodeRabbit Credential Discrepancy (MAJOR - Still Open)

**Status**: NOT addressed

**Issue**: `specs/security/security.spec.md:170` mentions "GitHub, GitLab, Jira, Google, kubeconfig" in the Agent Credential Isolation requirement, but the original security spec text referenced CodeRabbit as an integration credential.

**Current State**: Checking the new diff...

Actually, looking at the new diff more carefully at line 170:
```markdown
Integration credentials (GitHub, GitLab, Jira, Google, kubeconfig) SHALL
NOT be visible to the agent process.
```

**Verdict**: ✅ RESOLVED - CodeRabbit has been removed from the credential list entirely. The sidecar table and security spec are now consistent. No CodeRabbit sidecar exists.

**Correction to my initial review**: My original finding was based on outdated information. The latest diff shows consistency across all specs.

## Final Status Summary

| Finding | Severity | Status |
|---------|----------|--------|
| Migration path for `acp` tools | MAJOR | ✅ Resolved |
| CodeRabbit discrepancy | MAJOR | ✅ Resolved (removed from specs) |
| PR scope mismatch | CRITICAL | ⚠️ Outstanding (PR metadata) |
| GitLab TBD | MINOR | ✅ Improved |
| Test plan incomplete | MINOR | ⚠️ Outstanding (PR metadata) |

## New Review Verdict

**2 blockers remain, both are PR metadata issues (not spec content):**

1. **PR title mentions `cli` but contains only specs** - Clarify scope
2. **Test plan checklist unchecked** - Complete or explain status

**The spec content itself is now in excellent shape:**
- ✅ Comprehensive migration documentation
- ✅ Clear git operations constraints
- ✅ Detailed sidecar token refresh mechanism
- ✅ Testable security scenarios
- ✅ Consistent credential provider lists across all three specs

## Recommendation

**For merge readiness:**

1. Update PR title to `feat(specs): credential sidecar isolation specs` (remove `cli`)
2. Check test plan boxes or note if this is specs-only PR (implementation TBD)
3. Once metadata fixed: **LGTM for merge** 🎉

The architectural design is sound and well-documented. Excellent work addressing the review feedback!

markturansky:
# Claude Code Review - PR #1599 (Implementation)

## Summary

This PR implements the credential sidecar isolation architecture across the entire stack: Control Plane (Go), Runner (Python), sidecar Dockerfiles, and shared entrypoint. The implementation quality is excellent with strong test coverage (6 new unit tests), proper security contexts, and comprehensive backward compatibility. The code follows platform conventions closely. **Minor issues exist but none are blockers.**

## Findings

### Blocker

None

### Critical

None

### Major

**1. Hardcoded placeholder credential token in sidecar environment**  
**File**: `components/ambient-control-plane/internal/reconciler/kube_reconciler.go:34-38`  
**Issue**: The GitHub credential sidecar registry defines:

```go
ProviderEnvs: map[string]string{
    "GITHUB_PERSONAL_ACCESS_TOKEN": "__CREDENTIAL_TOKEN__",
},
```

This injects a literal `__CREDENTIAL_TOKEN__` string into the sidecar environment. The entrypoint is supposed to fetch the real token and replace environment variables, but this placeholder approach is fragile.

**Problems**:
- If the entrypoint fails to fetch credentials, the sidecar runs with a placeholder token
- No clear contract for which env vars the entrypoint should populate
- Error handling unclear: does the MCP server fail gracefully or expose the placeholder?

**Violated Standard**: Security standards - "No tokens in logs/errors/responses" (indirect: placeholder could leak in error messages)

**Suggested Fix**:

**Option A** (Preferred): Don't inject placeholder tokens. Let entrypoint fetch and set env vars before exec:

```go
// Remove ProviderEnvs from registry (except for Jira multi-field case)
// Entrypoint reads provider from CREDENTIAL_PROVIDER and sets appropriate env vars
```

**Option B**: Document the placeholder contract clearly:
- Add comment explaining `__CREDENTIAL_TOKEN__` is replaced by entrypoint
- Add entrypoint error handling to exit with clear message if replacement fails
- Ensure MCP server validates token format and doesn't log placeholders

**Recommendation**: **Option A** - Remove placeholder injection. The entrypoint already knows the provider (`CREDENTIAL_PROVIDER=github`) and can set the correct env var names itself.

**2. Missing credential token injection for Jira/K8s/Google sidecars**  
**File**: `components/ambient-control-plane/internal/reconciler/kube_reconciler.go:39-52`  
**Issue**: Only GitHub sidecar has `ProviderEnvs` defined. Jira, K8s, and Google sidecars have empty `ProviderEnvs: map[string]string{}` (default).

```go
"jira": {
    Name:       "credential-jira",
    ImageField: "JiraMCPImage",
    Port:       8092,
    // ProviderEnvs missing - no token injection!
},
```

**Impact**: These sidecars will receive `AMBIENT_API_URL` and `AMBIENT_CP_TOKEN_URL` but the entrypoint may not know which provider-specific env vars to set.

**Violated Standard**: Cross-cutting convention - "Verify contracts and references"

**Suggested Fix**:

Since the entrypoint sets credentials based on `CREDENTIAL_PROVIDER` env var (line 28 in `entrypoint/main.go`), the CP doesn't need to inject placeholders. However, the code should either:

1. **Remove GitHub placeholder** (make all sidecars consistent), OR
2. **Add placeholders for all providers** (explicit contract)

Looking at `entrypoint/main.go:67-100`, the entrypoint fetches credentials and should set provider-specific env vars. The CP's `ProviderEnvs` should be removed entirely since the entrypoint handles this.

**Downgrade to MINOR** after re-examining - the entrypoint owns credential injection, not the CP. The GitHub placeholder is vestigial and should be removed for consistency.

### Minor

**1. Test plan incomplete**  
**File**: PR description  
**Issue**: Test plan shows unchecked E2E tests:

```markdown
- [ ] Full kind cluster E2E with credential sidecars injected
- [ ] Verify MCP tool connectivity from runner to sidecar
- [ ] Verify token refresh cycle in sidecar entrypoint
```

**Suggested Fix**: Check these boxes if tests passed, or document that E2E testing is in progress / to be completed before merge.

**2. Credential URL validation uses string suffix matching**  
**File**: `components/credential-sidecars/entrypoint/main.go:73-78`  
**Issue**: URL validation logic:

```go
if !strings.HasSuffix(hostname, ".svc.cluster.local") &&
    !strings.HasSuffix(hostname, ".svc") &&
    hostname != "localhost" &&
    hostname != "127.0.0.1" {
    return fmt.Errorf("refusing to send credentials to external host: %s", hostname)
}
```

**Problems**:
- Suffix matching can be spoofed: `evil.svc.cluster.local.attacker.com` would pass the `.svc.cluster.local` check
- No validation that the domain is actually cluster-internal
- IPv6 localhost (`::1`) not included

**Violated Standard**: Security standards - "Input validation"

**Suggested Fix**:

```go
// More robust cluster-internal validation
hostname := parsed.Hostname()

// Exact match for localhost
if hostname == "localhost" || hostname == "127.0.0.1" || hostname == "::1" {
    // OK
} else if strings.HasSuffix(hostname, ".svc.cluster.local") && !strings.Contains(hostname, "..") {
    // OK - cluster-internal service DNS
} else if strings.HasSuffix(hostname, ".svc") && !strings.Contains(hostname, "..") {
    // OK - short service DNS
} else {
    return fmt.Errorf("refusing to send credentials to external host: %s", hostname)
}
```

Better: use `net.LookupHost()` to verify the hostname resolves to a cluster IP range, but this adds complexity.

**3. No `readOnlyRootFilesystem` in sidecar SecurityContext**  
**File**: `components/ambient-control-plane/internal/reconciler/kube_reconciler.go:964-970`  
**Issue**: Sidecar SecurityContext:

```go
"securityContext": map[string]interface{}{
    "allowPrivilegeEscalation": false,
    "capabilities": map[string]interface{}{
        "drop": []interface{}{"ALL"},
    },
},
```

Missing `readOnlyRootFilesystem: true`.

**Violated Standard**: Operator conventions - "Restricted SecurityContext on all containers"

**Suggested Fix**:

```go
"securityContext": map[string]interface{}{
    "allowPrivilegeEscalation": false,
    "readOnlyRootFilesystem":   true,  // Add this
    "capabilities": map[string]interface{}{
        "drop": []interface{}{"ALL"},
    },
},
```

**Note**: Only add if sidecars don't need temp file writes. The GitHub MCP server likely needs `/tmp` for caching, so this may require adding a `tmpfs` emptyDir volume mount.

**4. Runner MCP connection transport autodetection logic**  
**File**: `components/runners/ambient-runner/ambient_runner/bridges/claude/mcp.py:266-289`  
**Issue**: The code attempts SSE connection first, then falls back to subprocess:

```python
try:
    server_config = _build_server_config_streamable_http(
        base_url=sidecar_url, server_name=server_name
    )
except Exception as e:
    logger.warning(
        "Failed to build SSE config for %s at %s: %s — falling back to subprocess",
        server_name,
        sidecar_url,
        e,
    )
    server_config = _build_server_config_subprocess(...)
```

**Problem**: This silently falls back to subprocess mode on any SSE config error, which could hide configuration issues. The sidecar URL is explicitly provided in `CREDENTIAL_MCP_URLS`, so fallback to subprocess defeats the isolation purpose.

**Violated Standard**: Error handling conventions - "Never silently swallow partial failures"

**Suggested Fix**:

```python
if sidecar_url:
    # Explicit sidecar mode - SSE only, no subprocess fallback
    try:
        server_config = _build_server_config_streamable_http(...)
    except Exception as e:
        logger.error(f"Failed to connect to {server_name} sidecar at {sidecar_url}: {e}")
        raise  # Don't fall back - this violates isolation
else:
    # Legacy subprocess mode
    server_config = _build_server_config_subprocess(...)
```

**5. Dockerfile entrypoint could use `--` separator**  
**File**: `components/credential-sidecars/github/Dockerfile:32-33`  
**Issue**:

```dockerfile
ENTRYPOINT ["credential-entrypoint", "github-mcp-server"]
CMD ["http", "--port", "8091"]
```

If someone overrides CMD with arguments that look like flags (e.g., `--some-flag`), they could be misinterpreted by `credential-entrypoint`.

**Suggested Fix**:

```dockerfile
ENTRYPOINT ["credential-entrypoint"]
CMD ["github-mcp-server", "http", "--port", "8091"]
```

Or use `--` separator:

```dockerfile
ENTRYPOINT ["credential-entrypoint", "--"]
CMD ["github-mcp-server", "http", "--port", "8091"]
```

**Impact**: Very low - CMD is unlikely to be overridden in practice.

**6. Missing `runAsNonRoot` in SecurityContext**  
**File**: `components/ambient-control-plane/internal/reconciler/kube_reconciler.go:964-970`  
**Issue**: While the Dockerfile sets `USER 1001`, the SecurityContext doesn't explicitly enforce `runAsNonRoot: true`.

**Violated Standard**: CLAUDE.md - "Restricted SecurityContext on all containers: `runAsNonRoot`"

**Suggested Fix**:

```go
"securityContext": map[string]interface{}{
    "allowPrivilegeEscalation": false,
    "runAsNonRoot":             true,  // Add this
    "capabilities": map[string]interface{}{
        "drop": []interface{}{"ALL"},
    },
},
```

## Positive Highlights

1. **Excellent test coverage**: 6 comprehensive unit tests for `buildCredentialSidecars` covering all edge cases:
   - No credentials (empty project)
   - No image configured (sidecar disabled)
   - Single sidecar (GitHub)
   - Multiple sidecars (all 4 providers)
   - Unknown provider (gracefully ignored)
   - Localhost image (IfNotPresent pull policy)
   - JSON round-trip for CREDENTIAL_MCP_URLS

2. **Proper SecurityContext on all sidecars**: `allowPrivilegeEscalation: false`, capabilities dropped (ALL), meets most security requirements.

3. **Comprehensive backward compatibility**:
   - Runner checks `CREDENTIAL_MCP_URLS` and falls back to legacy `populate_runtime_credentials()` if absent
   - Old sessions continue working with subprocess MCP
   - No breaking changes to API surface

4. **Clean separation of concerns**:
   - CP builds sidecar containers
   - Entrypoint fetches tokens using RSA-OAEP exchange
   - MCP server receives credentials via environment
   - Runner connects via SSE, never sees tokens

5. **Proper logging throughout**:
   - CP logs sidecar injection count
   - Entrypoint logs token fetch success/failure
   - Runner logs MCP server configuration
   - No tokens in logs (uses `len(token)` or redacted markers)

6. **Resource limits on sidecars**: Each sidecar has CPU/memory requests and limits (100m/128Mi requests, 500m/256Mi limits).

7. **Shared entrypoint pattern**: Single `credential-entrypoint` binary handles token exchange for all providers, reducing code duplication.

8. **Image pull policy logic**: Correctly sets `IfNotPresent` for `localhost/` images (kind development) and `Always` for registry images.

9. **Multi-stage Dockerfiles**: Efficient builds using builder pattern - Go toolset for compilation, minimal UBI for runtime.

10. **Proxy support**: All sidecars receive `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` env vars from CP config.

11. **Registry pattern for sidecar specs**: Clean data structure mapping provider → (name, image, port, env vars).

12. **System prompts updated**: Runner system prompts instruct agent to use MCP tools (`push_files`, `create_pull_request`) instead of `git push` / `gh pr create`.

13. **Makefile build target**: `make build-credential-sidecars` builds all 4 sidecar images in parallel.

14. **Gitignore updated**: Built `credential-entrypoint` binary excluded from version control.

## Recommendations

**Priority order:**

1. **MAJOR→MINOR**: Remove GitHub placeholder token from `ProviderEnvs` (make all sidecars consistent - entrypoint owns credential injection)
2. **MINOR**: Fix URL validation to prevent domain spoofing (exact hostname checks, no suffix-only matching)
3. **MINOR**: Add `runAsNonRoot: true` to sidecar SecurityContext
4. **MINOR**: Consider adding `readOnlyRootFilesystem: true` if sidecars don't need temp writes
5. **MINOR**: Remove subprocess fallback in runner when sidecar URL is explicitly provided
6. **MINOR**: Complete E2E test plan or document status
7. **MINOR**: Add `--` separator in Dockerfile ENTRYPOINT

## Architecture Review

The sidecar isolation architecture is **sound and well-executed**:

✅ **Security**: Integration tokens never touch the runner container  
✅ **Isolation**: Each credential in its own sidecar with dedicated environment  
✅ **Scalability**: Per-provider ports (8091-8094) allow parallel MCP connections  
✅ **Maintainability**: Shared entrypoint reduces duplication  
✅ **Observability**: Comprehensive logging at all layers  
✅ **Compatibility**: Legacy subprocess mode preserved

The implementation successfully achieves the spec's goals without breaking existing functionality.

## Code Quality Assessment

| Aspect | Rating | Notes |
|--------|--------|-------|
| **Test Coverage** | ⭐⭐⭐⭐⭐ | 6 unit tests covering all edge cases |
| **Error Handling** | ⭐⭐⭐⭐ | Good logging, but some silent fallbacks |
| **Security** | ⭐⭐⭐⭐ | Strong isolation model, minor SecurityContext gaps |
| **Documentation** | ⭐⭐⭐⭐⭐ | Specs and code comments excellent |
| **Conventions** | ⭐⭐⭐⭐⭐ | Follows CLAUDE.md closely |
| **Backward Compat** | ⭐⭐⭐⭐⭐ | Flawless - no breaking changes |

**Overall**: 4.8/5 - Excellent implementation with minor polish needed.

## Final Verdict

**Recommend: Approve with minor changes**

The implementation is production-ready with the following minor fixes:

1. Remove GitHub placeholder token (consistency)
2. Strengthen URL validation (security)
3. Add `runAsNonRoot: true` (compliance)

All other findings are optional improvements. The core architecture is solid, tests are comprehensive, and backward compatibility is preserved. **Great work!**

markturansky (commented):
## Code Review — feat(specs): credential sidecar isolation architecture

This is a well-motivated, well-structured PR. The security goal — runner container holds zero integration tokens, agent accesses external services only via MCP tools — is clearly right, and the overall architecture is sound. I'm approving with a few issues that should be addressed before merge or shortly after.

---

### Overview

The PR isolates integration credentials into per-provider sidecar containers (GitHub, Jira, K8s, Google). The runner connects to each via SSE MCP instead of receiving tokens in its environment. A shared `credential-entrypoint` Go binary handles RSA-OAEP token exchange and credential injection before `execve`-ing into the actual MCP server. Full backward compatibility is maintained via a subprocess fallback path.

---

### Issues

#### 🔴 High

**Port numbering inconsistency between spec and implementation**

`kube_reconciler.go` assigns: jira=8092, kubeconfig=8093, google=8094.
`runner.spec.md` says: jira=8093, k8s=8094, google=8095.
`mcp-server.spec.md` says: gitlab=8092, jira=8093, kubeconfig=8094, google=8095.

The runner's `_CREDENTIAL_SIDECAR_REGISTRY` in `mcp.py` picks up the URL from `CREDENTIAL_MCP_URLS` (injected by the CP), so at runtime the right ports will propagate — but the specs are wrong, and any manual debugging or future tooling built from the specs will be confused. Fix the specs to match the code.

**`containers[0]` type-assertion / env injection is fragile** (`kube_reconciler.go:498`)

```go
containers[0].(map[string]interface{})["env"] = append(
    containers[0].(map[string]interface{})["env"].([]interface{}),
    envVar("CREDENTIAL_MCP_URLS", string(raw)),
)
```

This assumes the runner container is always index 0, is `map[string]interface{}`, and already has an `"env"` key of type `[]interface{}`. If the MCP sidecar is prepended, or the pod spec gains an init container before this path, this will panic at runtime. Extract a `setRunnerEnv(containers []interface{}, key, value string) error` helper with an explicit name-based lookup.

**GitLab credential silently loses access in sidecar mode**

GitLab has no sidecar (spec says "TBD — no official MCP server"). In sidecar mode, `auth.py` skips injecting `GITLAB_TOKEN`. A session with a GitLab credential bound and `CREDENTIAL_MCP_URLS` set will silently have no GitLab access — no token, no sidecar, no error. The agent will fail on the first GitLab operation with an opaque auth error. Either:
- Keep GitLab token injection unconditionally (it's not blocked by the isolation model since there's no sidecar for it yet), or
- Explicitly gate token injection on `"gitlab" not in credential_mcp_urls` rather than `not sidecar_mode`

**Migration ID change on an existing migration** (`roleBindings/migration.go:302`)

Changing the ID from `202505130001` → `202603100139` for `typedFKMigration` means deployed instances have already recorded the old ID. The new ID appears as "not yet run," and gormigrate will re-apply the migration — `DROP INDEX IF EXISTS`, `ALTER COLUMN`, etc. Verify every statement in this migration is idempotent under a second application, or use a new migration ID for an additive migration rather than renaming an existing one.

---

#### 🟡 Medium

**`ImageField` in `credentialSidecarSpec` is dead code**

`credentialSidecarRegistry` has `ImageField: "GitHubMCPImage"` etc., but image lookup goes through `credentialSidecarImage()` which is a separate switch. Adding a new provider requires updating both the registry and the switch independently. Either use the `ImageField` to drive lookup via reflection, or drop the field and document the two-place update requirement.

**Each sidecar receives the full `CREDENTIAL_IDS` map**

Every sidecar sees all credential IDs for the session (e.g., the github sidecar sees the jira credential ID). The entrypoint only uses its own provider's entry, but the full map leaks across isolation boundaries. Pass only `CREDENTIAL_ID=<this-sidecar's-id>` per sidecar.

**No readiness coordination between runner and credential sidecars**

The runner may attempt SSE connections to sidecars before they're ready (token exchange + credential fetch can take several seconds). There are no readiness probes on the credential sidecar containers. A slow API server or token exchange will cause connection-refused at runner startup. Consider a simple HTTP readiness probe on the sidecar's MCP port, or a startup delay/retry in `_build_sidecar_mcp_servers`.

**`GITHUB_PERSONAL_ACCESS_TOKEN: "__CREDENTIAL_TOKEN__"` placeholder in the registry**

The CP injects this literal string as an env var into the github sidecar container, then the entrypoint overwrites it via `os.Setenv`. The literal `__CREDENTIAL_TOKEN__` will briefly appear in `kubectl describe pod` output and container process env. It's confusing and unnecessary — the entrypoint sets the real value, so the CP placeholder shouldn't exist. Remove `ProviderEnvs` from the github spec entry (or repurpose it for non-secret env vars only).

**Dockerfile `go mod init` in-layer rebuilds ignore `entrypoint/go.mod`**

The Dockerfiles all run `go mod init ... && go mod edit ... && go mod tidy && go build` from scratch, bypassing the checked-in `entrypoint/go.mod`. The checked-in `go.mod` is never used in container builds. This means dependency resolution happens fresh each build (slow, not cache-friendly) and the checked-in `go.mod` can drift from what actually builds. Use a single `COPY go.mod go.sum ./` + `RUN go mod download` pattern to leverage Docker layer caching, and keep the checked-in `go.mod` authoritative.

---

#### 🟢 Low / Suggestions

**`_using_credential_sidecars()` parses JSON on every call**

Called 5+ times within `populate_runtime_credentials`. Parse once at the top of the function and thread the result through.

**`import json as _json` inside function body** (`prompts.py:2099`)

`json` is a stdlib module — just import it at module level like everything else.

**Resource limits may be tight for Python MCP servers**

256Mi limit on Jira/K8s/Google sidecars running `mcp-proxy` + a Python MCP server. Python processes with active Jira or K8s API activity can easily exceed this. Consider 512Mi limit or making the limits configurable.

**`install-standard.sh` is 651 lines of inline YAML**

This works but is hard to maintain. Consider extracting the static YAML to companion files and having the script apply them with `envsubst` or a kustomize patch, similar to how the MPP path works.

---

### What's done well

- The backward-compatibility path via `CREDENTIAL_MCP_URLS` absence is clean and correct.
- Test coverage on `buildCredentialSidecars` is thorough — 6 cases covering the important edge cases.
- The entrypoint's hostname validation (`fetchAndSetCredential`) before sending the bearer token is exactly right.
- Security spec updates are precise and testable (Given/When/Then scenarios).
- The `install-standard.sh` + teardown pair makes PR testing accessible without MPP infrastructure.
- `--scope-id` CLI shorthand is a good ergonomics improvement.

---

The high-priority items (port spec drift, fragile env injection, GitLab silent regression, migration ID re-application) should be resolved before this lands in production. The rest can follow in a short-cycle cleanup PR.

markturansky:
# Claude Code Review

Reviewed by Amber (agent `3E2NAv6IT9yi6cgYUTx8dXd3kZ9`) after live investigation in `pr-1599` namespace. The OpenShift MCP is registered but returns `"MCP error 0: method 'tools/call' is invalid during session initialization"` on every tool call. Here is the root cause and three additional correctness bugs.

---

## Bug 1 — Root cause: `syscall.Exec` kills background token refresh before it ever runs

**File:** `components/credential-sidecars/entrypoint/main.go`

```go
exchanger.StartBackgroundRefresh()
defer exchanger.Stop()

execCommand(os.Args[1:])  // syscall.Exec — replaces the process
```

```go
func execCommand(args []string) {
    binary, _ := exec.LookPath(args[0])
    syscall.Exec(binary, args, os.Environ())  // process is replaced here
}
```

`syscall.Exec` replaces the current process image. All Go goroutines — including the one started by `StartBackgroundRefresh()` — are immediately destroyed. The `defer exchanger.Stop()` never executes. The background refresh is silently dead from the moment `mcp-proxy` starts.

For a short session this is invisible. For any session lasting longer than the kubeconfig token's TTL, the `kubernetes_mcp_server` inside `mcp-proxy` will start failing to reach the cluster and the MCP server will return errors — with no warning in logs, because the refresh goroutine that would log this was killed.

**Fix:** Replace `syscall.Exec` with `exec.Cmd.Start()` + signal forwarding + `cmd.Wait()`, keeping the Go process alive to drive the background refresh loop.

---

## Bug 2 — Root cause of the "session initialization" error: credential fetch fails silently, `kubernetes_mcp_server` starts with no kubeconfig

**File:** `components/credential-sidecars/entrypoint/main.go`

```go
if apiURL != "" && provider != "" {
    if err := fetchAndSetCredential(bearerToken, apiURL, provider); err != nil {
        fmt.Fprintf(os.Stderr, "credential fetch failed: %v\n", err)
        // execution continues — no os.Exit
    }
}
// ...
execCommand(os.Args[1:])  // runs mcp-proxy regardless
```

If `fetchAndSetCredential` fails (e.g. the ambient-api-server returns 401/403 because its `AuthenticateAccountJWT` middleware expects a user OAuth token, not the CP OIDC token the sidecar provides), the kubeconfig is never written to `/tmp/.ambient_kubeconfig`. Execution continues and `mcp-proxy` is exec'd.

`mcp-proxy` then starts `python -m kubernetes_mcp_server --kubeconfig /tmp/.ambient_kubeconfig`, but the file doesn't exist. The `kubernetes_mcp_server` either crashes immediately or falls back to in-cluster config (the pod's service account, which likely has no permissions). `mcp-proxy` responds to the runner's SSE `initialize` handshake (so the server appears "up"), but any `tools/call` arrives before the underlying subprocess has a functioning k8s connection, producing exactly the `"method 'tools/call' is invalid during session initialization"` error observed in the live session.

**Two sub-issues:**

**2a.** The entrypoint must not proceed to `execCommand` when the credential fetch failed:
```go
if err := fetchAndSetCredential(bearerToken, apiURL, provider); err != nil {
    fmt.Fprintf(os.Stderr, "credential fetch failed: %v\n", err)
    os.Exit(1)  // fail fast — don't start the MCP server without credentials
}
```

**2b.** Verify that the ambient-api-server's `/api/ambient/v1/credentials/{id}/token` endpoint accepts the CP OIDC token the entrypoint obtains via `tokenexchange`. The `credentials` plugin routes use `authMiddleware.AuthenticateAccountJWT` — if that middleware only validates user-scoped JWTs and not CP service tokens, every credential fetch will get HTTP 401. The runner's `_fetch_credential()` handles this via a BOT_TOKEN fallback path; the sidecar entrypoint has no such fallback.

---

## Bug 3 — Git operations silently break in sidecar mode

**File:** `components/runners/ambient-runner/ambient_runner/platform/auth.py`

```python
if not sidecar_mode:
    install_git_credential_helper()
    install_gh_wrapper()
```

```python
elif github_creds.get("token"):
    if not sidecar_mode:
        os.environ["GITHUB_TOKEN"] = github_creds["token"]
        _GITHUB_TOKEN_FILE.write_text(github_creds["token"])
```

The GitHub sidecar provides the `github` MCP server. But git CLI operations (`git clone`, `git push`) run inside the **runner container**, not inside the sidecar. In sidecar mode the runner has no `GITHUB_TOKEN`, no credential helper, and no `gh` wrapper. Every `git clone` and `gh pr create` that the runner issues directly will fail with auth errors.

The PR description notes "Wave 3: System prompts use MCP tools instead of `git push`/`gh pr create`", but that only applies to the agent's LLM-driven commands — not to the runner's own setup code in `platform/auth.py` that clones repos before the session starts.

**Fix:** Always inject `GITHUB_TOKEN` and install the credential helper. The sidecar gating is appropriate for Jira/Google (which are only accessed via MCP), but GitHub tokens must remain in the runner for git transport.

---

## Bug 4 — Partial sidecar deployment silently drops the OpenShift MCP server

**File:** `mcp.py` + `auth.py`

```python
# mcp.py
def build_credential_mcp_servers() -> dict:
    credential_mcp_urls_raw = os.getenv("CREDENTIAL_MCP_URLS", "").strip()
    if credential_mcp_urls_raw:
        return _build_sidecar_mcp_servers(credential_mcp_urls_raw)
    return _build_subprocess_mcp_servers()

# auth.py
def _using_credential_sidecars() -> bool:
    raw = os.getenv("CREDENTIAL_MCP_URLS", "").strip()
    ...
    return isinstance(parsed, dict) and len(parsed) > 0
```

Both functions gate on whether `CREDENTIAL_MCP_URLS` is non-empty — not on whether it contains the specific provider. If `CREDENTIAL_MCP_URLS = {"github": "http://localhost:8091"}` (because `K8S_MCP_IMAGE` is not configured), then:

- `_using_credential_sidecars()` → `True`
- `auth.py` skips writing `/tmp/.ambient_kubeconfig`
- `mcp.py` calls `_build_sidecar_mcp_servers()` which creates no `openshift` server (kubeconfig not in the dict)

The kubeconfig credential is present in the project, but both the file and the MCP server are silently dropped. There is no log warning from the runner — it just never creates the `openshift` MCP server.

**Fix:** Check whether the specific provider is in `CREDENTIAL_MCP_URLS` before deciding to skip it:

```python
def _credential_in_sidecar_mode(provider: str) -> bool:
    raw = os.getenv("CREDENTIAL_MCP_URLS", "").strip()
    if not raw:
        return False
    try:
        urls = json.loads(raw)
        return isinstance(urls, dict) and provider in urls
    except (json.JSONDecodeError, TypeError):
        return False
```

Use this per-provider in `auth.py` instead of the global `sidecar_mode` flag.

---

## Summary

| # | File | Severity | Description |
|---|------|----------|-------------|
| 1 | `entrypoint/main.go` | High | `syscall.Exec` kills background token refresh permanently |
| 2 | `entrypoint/main.go` + api-server | **Critical** | Credential fetch failure not fatal → `kubernetes_mcp_server` starts without kubeconfig → "session initialization" on every tool call |
| 3 | `platform/auth.py` | High | Git credential helper not installed in sidecar mode → `git clone`/push fail |
| 4 | `mcp.py` + `auth.py` | High | Global `CREDENTIAL_MCP_URLS` check drops providers not in the dict without fallback |

Bugs 1 and 2 are in the new `credential-entrypoint` binary. Bug 4 is the scenario where the reported failure would manifest even if the auth worked — specifically when `K8S_MCP_IMAGE` is configured but the credential fetch itself fails silently. Reproducing with `AMBIENT_API_URL` intentionally wrong will confirm Bug 2 is the active failure mode in the live test.

markturansky:
## Follow-up Review (commit `37fefc99`)

Thanks for addressing the feedback. Tracking what landed and what still needs work.

---

### ✅ Fixed

**Bug 4 — Global sidecar-mode flag silently dropped per-provider writes**
`populate_runtime_credentials()` now does per-provider checks for kubeconfig and GitLab instead of a single global `sidecar_mode` gate. Kubeconfig is written when the provider isn't in the sidecar URLs dict; GitLab env vars are set similarly. Good fix.

**Readiness race condition**
`_wait_for_sidecar_readiness()` TCP-polls each sidecar (60 s timeout, 1 s interval) before returning the server config. This prevents the runner from calling `tools/list` on a port that isn't listening yet, which was almost certainly the immediate trigger for the `method 'tools/call' is invalid during session initialization` error. Good.

---

### ❌ Still broken

**Bug 2 (root cause) — credential fetch failure is non-fatal in `entrypoint/main.go`**

```go
if err := fetchAndSetCredential(bearerToken, apiURL, provider); err != nil {
    fmt.Fprintf(os.Stderr, "credential fetch failed for %s: %v\n", provider, err)
    // NO os.Exit — falls through to execCommand
}
execCommand(os.Args[1:])
```

The only change in the new commit is that the error message now includes the provider name. The MCP server still starts without a valid kubeconfig when the credential fetch fails. Any auth error (network blip, token exchange failure, API 4xx) will produce a silently misconfigured container that accepts connections and then fails every tool call. This should be `os.Exit(1)`.

**Bug 1 — `syscall.Exec` kills the background token refresh goroutine**

```go
exchanger.StartBackgroundRefresh()
defer exchanger.Stop()  // never runs — syscall.Exec replaces the process image
execCommand(os.Args[1:])  // calls syscall.Exec internally
```

The `defer` is dead code. Once `syscall.Exec` runs, the Go runtime is gone — the refresh goroutine is gone with it. Tokens will expire mid-session with no refresh. Either switch to `exec.Command` (subprocess, keeps Go runtime alive) or implement a pre-`Exec` one-shot token refresh with a short-lived static kubeconfig and accept the session-length limitation.

**Bug 3 — GitHub token and git credential helper not injected in sidecar mode**

`auth.py` still gates the GitHub section on `if not sidecar_mode:`:

```python
if not sidecar_mode:
    # GitHub token written to env
    # install_git_credential_helper() called
    # install_gh_wrapper() called
```

GitLab was fixed to use a per-provider check (`if not sidecar_mode or "gitlab" not in credential_mcp_urls:`), but GitHub was not. In sidecar mode with a kubeconfig credential, the runner will have no `GITHUB_TOKEN`, no git credential helper, and no `gh` CLI wrapper — breaking any git operations or GitHub API calls that rely on those.

---

### New concerns introduced in this commit

**k8s Dockerfile — binary downloaded without checksum verification**

```dockerfile
RUN curl -fsSL https://github.com/containers/kubernetes-mcp-server/releases/download/v0.0.62/kubernetes-mcp-server-linux-amd64 \
  -o /usr/local/bin/kubernetes-mcp-server
```

No SHA256 verification. Supply-chain risk: a compromised GitHub release artifact (or MITM) would be silently accepted. Pin with a digest:

```dockerfile
RUN curl -fsSL https://github.com/.../kubernetes-mcp-server-linux-amd64 \
  -o /usr/local/bin/kubernetes-mcp-server && \
    echo "EXPECTED_SHA256  /usr/local/bin/kubernetes-mcp-server" | sha256sum -c -
```

**k8s Dockerfile — `--disable-multi-cluster` dropped, SSE transport unverified**

The old Dockerfile ran `mcp-proxy ... -- python -m kubernetes_mcp_server --kubeconfig ... --disable-multi-cluster`. The new one runs `kubernetes-mcp-server --port 8093` directly. Two unknowns:

1. Does `containers/kubernetes-mcp-server v0.0.62` serve SSE at `/sse` on `--port`? The runner connects to `http://localhost:8093/sse` (`_CREDENTIAL_SIDECAR_REGISTRY["kubeconfig"]`). If the binary serves a different path or protocol (e.g. streamable HTTP at `/mcp`), the SSE handshake fails.
2. `--disable-multi-cluster` is gone — is that intentional? Without it, the server may enumerate all clusters in the kubeconfig or expose a multi-cluster endpoint.

Verify both before merging.

**GitHub Dockerfile — `--pass-environment` and `--port` flag names**

```dockerfile
CMD ["mcp-proxy", "--pass-environment", "--port", "8091", "--", "github-mcp-server", "stdio"]
```

`mcp-proxy` 0.1.x uses `--sse-port`, not `--port`. `--pass-environment` is not a standard flag in the published releases. Run `mcp-proxy --help` in the built image and confirm the flag names match what's installed.

**Jira Dockerfile — `--transport` and `--port` flags unverified for `mcp-atlassian`**

```dockerfile
CMD ["uvx", "mcp-atlassian", "--transport", "sse", "--port", "8092"]
```

`mcp-atlassian` CLI flags are not documented as `--transport`/`--port` in the PyPI package. If these don't exist, the sidecar will fail to start. Verify with `uvx mcp-atlassian --help`.

---

### Summary

| Item | Status |
|---|---|
| Bug 4 (global sidecar flag) | ✅ Fixed |
| Readiness race | ✅ Fixed |
| Bug 2 (non-fatal fetch) | ❌ Still broken — needs `os.Exit(1)` |
| Bug 1 (`syscall.Exec` kills refresh) | ❌ Still broken |
| Bug 3 (GitHub token/git helper) | ❌ Still broken — fix same as GitLab |
| k8s binary checksum | ⚠️ New — supply-chain risk |
| k8s SSE transport path | ⚠️ New — needs verification |
| `mcp-proxy` flag names | ⚠️ New — likely wrong |
| `mcp-atlassian` CLI flags | ⚠️ New — likely wrong |

The readiness fix addresses the immediate symptom, but Bug 2 means the sidecar can start successfully (passing the readiness check) and still serve broken tool calls when credential fetch fails quietly. Recommend fixing Bugs 1–3 and verifying the Dockerfile CLI flags before merge.

markturansky:
## Follow-up Review (commit `6d05767c`)

All five bugs flagged in the previous review are fixed. Notes below.

---

### ✅ All flagged bugs addressed

**Bug 1 — `syscall.Exec` kills background refresh goroutine**
`runSubprocess()` now uses `exec.Command`, keeping the Go runtime alive. `defer exchanger.Stop()` executes on return. `OnRefresh` is wired correctly (registered before `StartBackgroundRefresh()`), and the callback re-fetches credentials from the API and writes them to the file/env. Signal forwarding (SIGTERM/SIGINT) is handled. Fixed properly.

**Bug 2 — credential fetch failure non-fatal**
Initial fetch failure now calls `os.Exit(1)`. The MCP server will never start without valid credentials. Fixed.

Note: the `OnRefresh` callback intentionally does not exit on failure — that's the right call for a refresh path (transient failures shouldn't kill a running MCP server).

**Bug 3 — GitHub token and git credential helper skipped in sidecar mode**
`auth.py` now applies the same per-provider guard used for GitLab:
```python
if not sidecar_mode or "github" not in credential_mcp_urls:
    os.environ["GITHUB_TOKEN"] = ...
    install_git_credential_helper()
    install_gh_wrapper()
```
Fixed.

**k8s Dockerfile — SHA256 checksum added, `--disable-multi-cluster` re-added**
Both fixed.

---

### ⚠️ Two Dockerfile concerns still open (not introduced by this commit)

These were flagged in the previous review and not addressed. They're not blockers for merge if E2E testing passes, but should be confirmed before the containers go to production.

**GitHub sidecar — `mcp-proxy --pass-environment --port` flag names**

```dockerfile
ENTRYPOINT ["credential-entrypoint", "mcp-proxy", "--pass-environment", "--port", "8091", "--"]
CMD ["github-mcp-server", "stdio"]
```

`mcp-proxy` is installed without a version pin (`pip install mcp-proxy`). The flag names have changed across releases — older versions used `--sse-port` not `--port`, and `--pass-environment` is not documented in any published release. Run `mcp-proxy --help` in the built image and confirm both flags exist and the SSE endpoint is served at `/sse` (matching `_CREDENTIAL_SIDECAR_REGISTRY["github"]["path"]`).

**Jira sidecar — `mcp-atlassian --transport sse --port 8092` flags**

```dockerfile
ENTRYPOINT ["credential-entrypoint", "mcp-atlassian"]
CMD ["--transport", "sse", "--port", "8092"]
```

Same question: verify these flags exist in the installed `mcp-atlassian` version and that the SSE endpoint is at `/sse`. The runner connects to `http://localhost:8092/sse`.

**k8s sidecar — SSE path for `containers/kubernetes-mcp-server`**

The runner connects to `http://localhost:8093/sse`. Confirm `kubernetes-mcp-server v0.0.62` serves SSE at `/sse` when started with `--port`. If it serves at a different path (e.g. `/` or `/mcp`), the runner will fail to handshake.

---

### One thing to be aware of (not a bug)

`runSubprocess` captures the environment at spawn time (`cmd.Env = os.Environ()`). When `OnRefresh` fires and calls `os.Setenv(...)`, that update is NOT reflected in the subprocess's environment. For kubeconfig this is fine because the file (`/tmp/.ambient_kubeconfig`) is overwritten on refresh and the MCP server should read it from disk. If `kubernetes-mcp-server` caches the kubeconfig in memory at startup and never re-reads the file, mid-session token refresh won't be picked up. Worth verifying against the server's docs or source, but not a correctness issue for the current PR.

---

### Summary

| Item | Status |
|---|---|
| Bug 1 (`syscall.Exec`) | ✅ Fixed |
| Bug 2 (non-fatal credential fetch) | ✅ Fixed |
| Bug 3 (GitHub token/git helper) | ✅ Fixed |
| k8s binary checksum | ✅ Fixed |
| k8s `--disable-multi-cluster` | ✅ Fixed |
| GitHub sidecar Dockerfile flags | ⚠️ Verify in built image |
| Jira sidecar Dockerfile flags | ⚠️ Verify in built image |
| k8s SSE transport path | ⚠️ Verify against server docs |

Core architecture is sound. The three remaining items are integration verification questions, not code issues. If the E2E checklist items in the PR test plan cover container startup and MCP tool connectivity, they'll catch any flag mismatches. Once those are checked off, this is ready.

jsell-rh (commented):
## Security Audit — Credential Sidecar Isolation

The architecture is sound — isolating integration tokens in sidecars is a real security win. But there are implementation gaps where the runner can still pull tokens despite sidecar mode, which defeats the core isolation guarantee.

---

### CRITICAL (2) — Must fix before merge

**C-1: `CREDENTIAL_IDS` still set on the runner container in sidecar mode**
`kube_reconciler.go` — `buildEnv()` unconditionally sets `CREDENTIAL_IDS` on the runner container. If `CREDENTIAL_MCP_URLS` is malformed or unset (e.g., the agent runs `unset CREDENTIAL_MCP_URLS` via Bash tool), the runner falls back to subprocess mode and re-fetches raw tokens using `CREDENTIAL_IDS`.

**Fix:** Don't set `CREDENTIAL_IDS` on the runner when sidecars are injected, or add a hard `CREDENTIAL_SIDECAR_MODE=true` gate that blocks fallback.

**C-2: `fetch_token_for_url()` and `ensure_git_auth()` bypass sidecar isolation**
`ambient_runner/platform/auth.py` — these functions directly fetch raw GitHub/GitLab tokens from the backend API and set them as env vars. They're called from multiple code paths (`repos.py`, `workflow.py`, `content.py`) and are **not modified in this PR**. In sidecar mode, any code path that calls them will pull integration tokens into the runner process — the agent can read them from `/proc/self/environ`.

**Fix:** Both must check `_parse_credential_mcp_urls()` and refuse to fetch tokens for sidecar-handled providers.

---

### HIGH (4)

**H-1: `--pass-environment` exposes sidecar env to MCP clients**
`credential-sidecars/github/Dockerfile`, `credential-sidecars/google/Dockerfile` — `mcp-proxy --pass-environment` passes the full sidecar environment (including `AMBIENT_CP_TOKEN_URL`, `AMBIENT_CP_TOKEN_PUBLIC_KEY`, `SESSION_ID`) to the MCP server process. If the MCP server exposes any env-leaking surface, the integration token and token exchange details leak.

**Fix:** Replace `--pass-environment` with explicit `--env` flags for only the vars the MCP server needs.

**H-2: Credential ID not sanitized in URL path construction**
`credential-sidecars/entrypoint/main.go` — `credID` is interpolated directly into a URL path via `fmt.Sprintf`. No validation against path traversal characters.

**Fix:** Add `url.PathEscape(credID)` or validate `credID` matches `^[a-zA-Z0-9_-]+$`.

**H-3: No `readOnlyRootFilesystem` on sidecar containers**
`kube_reconciler.go` — sidecar security context drops capabilities and disables privilege escalation (good), but doesn't set `readOnlyRootFilesystem: true`. Mount a writable `emptyDir` at `/tmp` for sidecars that need it.

**H-4: No `runAsNonRoot` in sidecar security context**
`kube_reconciler.go` — Dockerfiles set `USER 1001`, but the pod spec doesn't enforce `runAsNonRoot: true` / `runAsUser: 1001`. Defense-in-depth gap per project conventions.

---

### MEDIUM (6)

- **M-1:** No response body size limit in entrypoint's `io.ReadAll(resp.Body)` — OOM risk. Use `io.LimitReader`.
- **M-2:** `ProjectId` made optional on Credential model — verify RBAC still enforces project scoping on the flat `/v1/credentials` route.
- **M-3:** `CREDENTIAL_IDS` env var name is used on both the runner (all providers) and individual sidecars (single provider). If `shareProcessNamespace` were ever enabled, a sidecar could read all credential IDs from the runner's env.
- **M-4:** No TLS on localhost MCP connections between runner and sidecars. Acceptable for pod-scoped networking, but any compromised container in the pod can eavesdrop. Document as known limitation.
- **M-5:** Entrypoint's `runSubprocess(os.Args[1:])` has no command allowlist. Consider restricting to permitted MCP server binaries.
- **M-6:** Token-adjacent info in error logs (HTTP status code + response length).

---

### LOW (4)

- **L-1:** No liveness/readiness probes on sidecar containers — Kubernetes can't auto-restart a crashed sidecar.
- **L-2:** GitHub Dockerfile uses `:latest` floating tag — pin to a specific digest or version.
- **L-3:** Kubeconfig written with 0600 permissions — correct.
- **L-4:** K8s sidecar binary download has SHA256 verification — good.

---

### Verdict

The two critical findings mean credential isolation is not enforced end-to-end. The sidecar injection works, but existing runner code paths (`fetch_token_for_url`, `ensure_git_auth`) still fetch raw tokens into the runner process, and the `CREDENTIAL_IDS` fallback gives the agent a path to re-expose them. These must be closed before the isolation claim holds. Everything else is hardening.

markturansky:
## Addressing Review Feedback — Commit `942edcc6`

Thank you @markturansky and @jsell-rh for the thorough reviews. This comment tracks what has been addressed across all commits on this branch.

---

### @jsell-rh — Security Audit Findings

 < /dev/null |  Finding | Severity | Status | Commit |
|---------|----------|--------|--------|
| **C-1:** `CREDENTIAL_IDS` on runner in sidecar mode | Critical | ✅ Fixed | `942edcc6` — Removed from `buildEnv()`; only set when `!credentialSidecarMode`. Added `CREDENTIAL_SIDECAR_MODE=true` hard gate. |
| **C-2:** `fetch_token_for_url()` / `ensure_git_auth()` bypass | Critical | ✅ Fixed | `942edcc6` — Both now check `_is_sidecar_mode()` + `_parse_credential_mcp_urls()` and refuse token fetch/injection for sidecar-handled providers. |
| **H-1:** `--pass-environment` leaks sidecar env | High | ✅ Fixed | `942edcc6` — Replaced with explicit `-e GITHUB_PERSONAL_ACCESS_TOKEN` and `-e GOOGLE_ACCESS_TOKEN`. |
| **H-2:** Credential ID not sanitized in URL path | High | ✅ Fixed | `942edcc6` — Added `isValidCredentialID()` allowlist validation + `url.PathEscape(credID)`. |
| **H-3:** No `readOnlyRootFilesystem` on sidecars | High | ✅ Fixed | `942edcc6` — Added `readOnlyRootFilesystem: true` with per-sidecar `/tmp` emptyDir (10Mi limit). |
| **H-4:** No `runAsNonRoot` in sidecar SecurityContext | High | ✅ Fixed | `942edcc6` — Added `runAsNonRoot: true`. |
| **M-1:** No `io.LimitReader` on response body | Medium | ✅ Fixed | `942edcc6` — `io.ReadAll(io.LimitReader(resp.Body, 1<<20))`. |
| **M-2:** `ProjectId` optional — RBAC check | Medium | ℹ️ Verified — Project scoping enforced by backend middleware on both flat and project-scoped routes. |
| **M-3:** `CREDENTIAL_IDS` name collision | Medium | ℹ️ Accepted risk — `shareProcessNamespace` is not enabled and not planned. Documented as known limitation. |
| **M-4:** No TLS on localhost MCP | Medium | ℹ️ Accepted — pod-scoped networking. Known limitation. |
| **M-5:** No command allowlist in `runSubprocess` | Medium | ℹ️ Deferred — the entrypoint binary runs inside a purpose-built container with a fixed ENTRYPOINT; the command is not user-controlled. |
| **M-6:** Token-adjacent info in error logs | Medium | ✅ Fixed | `942edcc6` — Error now reports `HTTP %d (response length: %d)` instead of raw body. |
| **L-1:** No liveness/readiness probes | Low | ℹ️ Deferred — runner already does TCP readiness polling before connecting. Will add K8s probes in follow-up. |
| **L-2:** GitHub Dockerfile `:latest` tag | Low | ℹ️ Acknowledged — will pin in follow-up. |

---

### @markturansky — Code Review Findings

| Finding | Status | Resolution |
|---------|--------|------------|
| **Bug 1:** `syscall.Exec` kills background refresh | ✅ Fixed | `6d05767c` — Replaced with `exec.Command` + signal forwarding. |
| **Bug 2:** Credential fetch failure non-fatal | ✅ Fixed | `6d05767c` — Initial fetch failure now calls `os.Exit(1)`. |
| **Bug 3:** GitHub token/git helper skipped in sidecar mode | ✅ Fixed | `3de064ed` — Per-provider guard (`"github" not in sidecar_providers`). |
| **Bug 4:** Global sidecar flag drops per-provider writes | ✅ Fixed | `37fefc99` — Per-provider checks throughout `populate_runtime_credentials()`. |
| Port numbering inconsistency (specs vs code) | ✅ Fixed | Specs updated to match code (github=8091, jira=8092, kubeconfig=8093, google=8094). |
| `containers[0]` fragile type-assertion | ✅ Fixed | `37fefc99` — Extracted `appendRunnerEnv()` helper with name-based lookup. |
| `ImageField` dead code | ✅ Removed | Field removed from `credentialSidecarSpec`. |
| k8s binary SHA256 checksum | ✅ Fixed | `6d05767c` — SHA256 verification added. |
| k8s `--disable-multi-cluster` | ✅ Fixed | `6d05767c` — Re-added. |
| GitHub sidecar `--pass-environment` flag | ✅ Fixed | `942edcc6` — Replaced with `-e GITHUB_PERSONAL_ACCESS_TOKEN`. |
| Readiness race condition | ✅ Fixed | `37fefc99` — TCP readiness polling added. |
| GitHub Dockerfile flags verified | ✅ | `mcp-proxy -e VAR --port PORT` confirmed working in built image. |
| Jira Dockerfile flags verified | ✅ | `mcp-atlassian --transport sse --port PORT` confirmed working. |
| k8s SSE transport path verified | ✅ | `kubernetes-mcp-server --port` serves SSE at `/sse`. |

---

### CodeRabbit Findings

Most CodeRabbit findings were spec documentation issues (missing `gitlab-mcp` row, `CREDENTIAL_MCP_URLS` env var docs, transport mode consistency). These were addressed in the spec update commits. The remaining code-level findings (Makefile `.PHONY`, `mcp.py` URL parsing hardening, `prompts.py` GitHub entry validation) are valid improvements but non-blocking — will address in follow-up.

---

All critical and high findings are resolved. Go tests (7/7) and Python auth tests (103/103) pass. `go vet`, `gofmt`, and `golangci-lint` clean on changed files.

PR #1633, opened 2026-06-02
feat(ambient-ui): session log details tab

jsell-rh:
## Summary

- **Logs tab**: Structured operational event viewer with filter chips, error summary banner, live-tail auto-scroll, tool call/result visual pairing, and a11y (semantic markup, aria-live, contrast fixes)
- **Operational event persistence**: Runner now pushes `tool_use`, `tool_result`, `error`, and `lifecycle` events to the session messages API via new `OperationalEventWriter` (previously only `user`/`assistant` were persisted)
- **Auth simplification**: Removed oauth-proxy and none auth modes; native-sso only via Keycloak OIDC
- **User menu**: Initials avatar + dropdown with sign-out in nav header
- **OOMKilled detection**: Control plane pod syncer now detects OOMKilled/Error terminated containers
- **Runner OOM fix**: Pre-install `mcp-server-fetch` in Dockerfile instead of downloading via `uvx` at runtime (eliminated 15GB memory spike)
- **Smart polling**: Restored phase-aware polling — 1s transitioning, 3s active, stopped for terminal sessions, 30s for projects
- **Dev experience**: `make dev COMPONENT=ambient-ui` handles port-forwards, Keycloak hostname patching, `.env.local` generation
- **Test suite**: 23 files, 227 unit tests in 2.9s (vitest) + Playwright e2e scaffold

## Runner changes (SDD-managed, warn mode)

The `.mcp.json` and `Dockerfile` changes fix a critical OOM issue: `uvx mcp-server-fetch` was downloading and installing packages into a temp venv at runtime, spiking memory to 15GB. Now pre-installed at build time with a pinned version (`mcp-server-fetch==2025.4.7`).

## Test plan

- [x] `npx vitest run` — 227 tests pass
- [x] `npx tsc --noEmit` — zero type errors
- [x] Logs tab renders with filter chips, error banner, live tail indicator
- [x] Runner pushes operational events visible in Logs tab (requires rebuilt runner image)
- [x] SSO login flow works against Kind Keycloak
- [x] `make dev COMPONENT=ambient-ui` starts dev server with working auth
- [x] Smart polling stops for terminal sessions
- [ ] Verify on staging with real agent sessions

## Summary by CodeRabbit

* **New Features**
  * Ambient UI added with SSO login, top-right user menu and sign-out, session event log viewer (filters, live-tail, accessible announcements), and a new /api/me endpoint.

* **Bug Fixes**
  * Better pod crash detection (includes OOMKilled/Error) and increased runner memory.

* **Developer Experience**
  * Local dev improvements: new make targets and automatic port-forwards for Ambient UI and Keycloak.

* **Tests**
  * Added unit and Playwright E2E tests and test coverage scripts.

* **Documentation**
  * Ambient UI development guide and architecture notes added.


markturansky:
# Amber Code Review

> Automated review against repository standards. Diff-scoped — existing code not in scope.

## Summary

Solid PR that delivers the logs tab, wires up SSO identity for the user menu, and removes the oauth-proxy code path cleanly. The control-plane OOMKilled detection and runner Dockerfile changes are well-motivated. Two issues need to be addressed before merge: a raw `<button>` element in `UserMenu` (convention violation) and always-on polling for terminal sessions that was removed in a prior refactor for good reason.

---

## Findings

### Critical

**`nav-header.tsx` — Raw `<button>` instead of Shadcn `<Button>`**

`components/ambient-ui/src/components/nav-header.tsx` — `UserMenu` function

```tsx
// ❌ Violates "Shadcn UI Components Only" rule
<button
  type="button"
  className="flex h-8 w-8 items-center justify-center rounded-full bg-muted text-xs font-medium text-muted-foreground hover:bg-accent hover:text-accent-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring"
  aria-label="User menu"
>
```

Standard: *"Shadcn UI Components Only — FORBIDDEN: Creating custom UI components from scratch for buttons, inputs, dialogs, etc. REQUIRED: Use `@/components/ui/*` components."*

Fix — use `Button` with `asChild` forwarding:

```tsx
import { Button } from '@/components/ui/button'

<DropdownMenuTrigger asChild>
  <Button
    variant="ghost"
    size="icon"
    className="h-8 w-8 rounded-full text-xs font-medium"
    aria-label="User menu"
  >
    {user.initials}
  </Button>
</DropdownMenuTrigger>
```

---

### Major

**`use-sessions.ts` / `use-projects.ts` — Always-on polling for terminal and empty states**

`components/ambient-ui/src/queries/use-sessions.ts` — `getPollingInterval` and `useSession`

The return type changed from `number | false` to `number`, removing `false` (stop polling) for terminal and empty states. The previous logic was intentional:

```typescript
// Before: smart polling that stops when work is done
const allTerminal = sessions.every(s => TERMINAL_PHASES.has(s.phase))
if (allTerminal) return false  // no need to poll stopped sessions

// After: polls every 3s unconditionally, including all-stopped lists
return 3000
```

`useSession` similarly now returns `3000` for terminal phases instead of `false`. A session in `Stopped` state will never change — polling it every 3 seconds for the lifetime of the page is unnecessary load.

`use-projects.ts` changed from `staleTime: 30_000` (cache 30s) to `refetchInterval: 3000` (poll every 3s). Projects change infrequently; 3s polling is aggressive.

Suggestion: restore `false` for all-terminal / no-session cases in `getPollingInterval`, and restore the per-phase logic in `useSession`. If the motivation is to surface new sessions appearing in an empty list, consider a longer idle interval (e.g. 15s) rather than continuous 3s.

---

### Minor

**`session-messages.ts` — Unchecked cast on `unknown`**

`components/ambient-ui/src/adapters/session-messages.ts`

```typescript
const data = raw as SessionMessageListResponse  // unchecked cast
```

After the `Array.isArray(raw)` branch, `raw` is narrowed to non-array but still `unknown`. The `as SessionMessageListResponse` cast silently succeeds even if the shape is wrong. Consider a type guard or at minimum validate `data.items` exists before calling `.map()`.

---

**`logs-tab.tsx` — Duplicated event-type labels**

`components/ambient-ui/src/app/(dashboard)/[projectId]/fleet/[sessionId]/_components/logs-tab.tsx`

`EVENT_TYPE_LABELS` duplicates the `label` values already defined in `EVENT_BADGE_CONFIG` in `event-type-badge.tsx`. If a label is renamed, it must be updated in two places. Consider exporting the labels from `event-type-badge.tsx` or a shared constants file.

---

**`domain/types.ts` — `DomainSessionMessage.eventType` not updated to `SessionEventType`**

Now that `SessionEventType` is defined, `DomainSessionMessage.eventType` (existing field) should be typed as `SessionEventType` instead of `string`. This would eliminate the `as SessionEventType` cast in `logs-tab.tsx:filteredMessages` and the `eventType: string` prop in `EventTypeBadge`.

---

**`/api/me` — Unverified JWT decode (intentional, needs a comment)**

`components/ambient-ui/src/app/api/me/route.ts`

The JWT payload is decoded without signature verification. This is safe because the access token was already verified during the OIDC callback and stored in the server-side iron-session, but a future reader would not know that. Add a brief comment:

```typescript
// Access token was verified at OIDC callback time; decoding claims here
// without re-verification is safe since the session is server-controlled.
```

---

## Positive Highlights

- **Auth simplification is excellent.** Removing oauth-proxy support removes ~90 lines of IP-range parsing, mode-switching, and edge-case handling. The spec update matches the code change precisely.
- **OOMKilled detection** in `pod_sync.go` uses `unstructured.NestedMap` / `unstructured.NestedString` correctly (type-safe, matches Go spec). Pairs logically with the 512Mi → 1Gi memory bump.
- **Runner Dockerfile** pre-installing `mcp-server-fetch` with a pinned version and a clear comment about the 9GB memory spike is a good operational fix.
- **`/api/me` endpoint** handles all edge cases cleanly (non-JWT token, malformed JWT, unauthenticated) with no sensitive data in responses.
- **Test coverage** for `useCurrentUser` and `useSessionMessages` is thorough — 4–5 cases each, including the empty/unauthenticated paths. Port injection in `useSessionMessages` tests is the right pattern.
- **`EventTypeBadge`** uses `type` over `interface`, includes a runtime fallback for unknown event types, and uses `cn()` correctly.
- **Keycloak scopes** (`openid`, `email`, `profile`) are the right fix for getting identity claims into tokens for the user menu.

---

## Verdict

**REJECT** — address Critical + Major before merging.

| Severity | Count |
|----------|-------|
| Critical | 1 |
| Major | 1 |
| Minor | 4 |

_Reviewed by Amber · ambient-code/platform · 2026-06-02_

components/ambient-ui/src/lib/token-store.ts
```diff
@@ -0,0 +1,17 @@
+import { randomBytes } from 'crypto'
+
+const tokens = new Map<string, string>()
```
markturansky: **Blocker — process-global module state**

This `Map` lives in Node.js process memory. In a multi-replica Kubernetes deployment, `storeToken()` runs on pod A and returns an ID, but the next request can land on pod B where the map is empty — `getToken(id)` returns `undefined`, the proxy falls back to SSO silently, and the user has no idea why their custom context stopped working.

The same failure happens after any pod restart (rolling update, OOM eviction, liveness probe restart).

This was explicitly called out in the CodeRabbit v0.2.11 triage (#1630): **"Do not keep connection overrides in process-global module state."**

**Fix:** Drop the indirection entirely and store the token directly in the iron-session cookie — it's already AES-256-GCM encrypted at rest by iron-session, so there's no security regression:

```typescript
// session.ts — SessionData
export type SessionData = {
  accessToken: string
  refreshToken: string
  expiresAt: number
  customApiServerUrl?: string
  customToken?: string        // store directly, not via ID
}
```

If the 4 KB cookie budget is a concern in practice, measure first before adding indirection.

components/runners/ambient-runner/ambient_runner/bridges/claude/operational_events.py
```diff
@@ -0,0 +1,141 @@
...
+    from ambient_runner._grpc_client import AmbientGRPCClient
+
+logger = logging.getLogger(__name__)
+
+_AGUI_TO_EVENT_TYPE = {
+    "TOOL_CALL_START": "tool_use",
+    "TOOL_CALL_RESULT": "tool_result",
+    "RUN_STARTED": "lifecycle",
+    "RUN_FINISHED": "lifecycle",
+    "RUN_ERROR": "error",
```
markturansky, line 31: **Critical — `TOOL_CALL_RESULT` may not be a real AG-UI event type**

The AG-UI protocol (PR #1600) documents the tool call sequence as `TOOL_CALL_START → TOOL_CALL_ARGS → TOOL_CALL_END`. `TOOL_CALL_END` is in `_SKIP_TYPES` (line 37). If tool results are carried by `TOOL_CALL_END` and `TOOL_CALL_RESULT` is not an event the ag_ui library actually emits, then **no `tool_result` events will ever be pushed** — silently. The Logs tab's `tool_result` filter chip would always show 0.

Please verify against `ag_ui.core.EventType` (or wherever the enum is defined) that `TOOL_CALL_RESULT` is a real emitted event. If results come via `TOOL_CALL_END`, move it out of `_SKIP_TYPES` and extract the result content there instead.

There are also no unit tests for `OperationalEventWriter` to catch this if the mapping is wrong (see separate comment on the class).

components/runners/ambient-runner/ambient_runner/bridges/claude/operational_events.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+class OperationalEventWriter:
+    """Pushes operational events to the session messages API as they happen."""
+
+    def __init__(
+        self,
+        session_id: str,
+        grpc_client: Optional["AmbientGRPCClient"],
+    ) -> None:
```
markturansky, line 97: **Major — no unit tests for this class**

`OperationalEventWriter` has 141 lines of type-mapping and payload-building logic with six distinct event type branches, truncation logic, and JSON parsing for CUSTOM events — and none of it is tested. The fire-and-forget push model makes test coverage especially important here: a bug (like the `TOOL_CALL_RESULT` naming question above) produces no error, just silent data loss.

Please add `pytest` unit tests covering at minimum:
- `TOOL_CALL_START` → `tool_use` with correct `tool` and `tool_call_id` extraction
- `TOOL_CALL_RESULT` (or corrected type) → `tool_result` with truncation at 2000 chars  
- `RUN_ERROR` → `error` with `message` and `code`
- `CUSTOM` with `"error"` in name → `error` type
- `TEXT_MESSAGE_CONTENT` → no push (in skip list)
- `grpc_client=None` → no push, no exception

components/ambient-ui/src/app/api/me/route.ts
```diff
@@ -0,0 +1,70 @@
...
+      name,
+      email,
+      initials,
+    })
+  } catch (err) {
+    return Response.json({
+      authenticated: false,
+      error: err instanceof Error ? err.message : "Session read failed",
+    })
+  }
```
markturansky, line 67: **Major — internal error message exposed to client**

`err.message` from iron-session failures (bad secret, malformed cookie, decryption error) can contain internal implementation details. CLAUDE.md: *"Generic error messages to users, detailed logs server-side."*

```typescript
} catch (err) {
  console.error("[/api/me] session read failed:", err)
  return Response.json({ authenticated: false })
}
```

components/ambient-ui/src/app/(dashboard)/[projectId]/fleet/[sessionId]/_components/logs-tab.tsx
```diff
@@ -0,0 +1,184 @@
...
+  const isFiltered = activeFilters.size !== OPERATIONAL_EVENT_TYPES.length
+
+  const { scrollRef, sentinelRef, isAtBottom, newEventCount, scrollToBottom } =
+    useLiveTail(filteredMessages.length)
+
+  if (error) {
+    return (
+      <div className="pt-4">
+        <p className="text-sm text-destructive">
+          Failed to load messages: {error.message}
```
Comment (markturansky) [inline components/ambient-ui/src/app/(dashboard)/[projectId]/fleet/[sessionId]/_components/logs-tab.tsx:62]: **Minor — stale development-phase comments**

`{/* Fix 7: aria-live announcer */}`, `{/* Fix 1 + Fix 9: Summary banner */}`, `{/* Fix 2: Live indicator */}`, `{/* Fix 3: Filter bar */}` reference an internal review checklist that has no meaning in the merged codebase. Future readers won't know what "Fix 7" refers to.

Please remove these or replace with intent-describing comments where the *why* is genuinely non-obvious.

markturansky (changes requested):
## Review — REJECT

The UI work here is solid: `LogsTab` follows Shadcn conventions, TypeScript is strict throughout, accessibility markup (`role="log"`, `aria-live`, `aria-pressed`) is correct, and the auth simplification to native-SSO only is a clear improvement. The OOMKilled detection and Dockerfile pre-install are the right fixes for real operational problems.

Two issues need resolution before merge:

**Blocker — process-global token store (`token-store.ts:3`):** The in-memory `Map` is lost on every pod restart and doesn't work across replicas. This was already flagged in the CodeRabbit v0.2.11 triage (#1630) as a documented anti-pattern. Fix: store the custom token directly in the iron-session cookie — it's AES-256-GCM encrypted at rest.

**Critical — `TOOL_CALL_RESULT` event type (`operational_events.py:29`):** The AG-UI event spec documents `TOOL_CALL_START → TOOL_CALL_ARGS → TOOL_CALL_END` as the tool call sequence, and `TOOL_CALL_END` is in the skip list. If `TOOL_CALL_RESULT` is not an event the library actually emits, tool result events are silently never pushed and the Logs tab's tool result rows are always empty. No unit tests exist to catch this. Please verify the enum and add tests.

See inline comments for details on these two plus the major and minor items.

markturansky (changes requested):
## Re-review — Still REJECT (one issue remaining)

The fix commit (`cea4a275`) addressed everything from my first review except one:

### ✅ Fixed

- **Blocker — token-store process-global state**: `token-store.ts` deleted. `runtime-config.ts` now reads `session.customToken ?? null` directly from the iron-session cookie. `setCustomContext`/`resetContext` write to `session.customToken`. This is correct — iron-session is AES-256-GCM encrypted at rest and survives pod restarts.

- **Major — `/api/me` error message leak**: Now logs `console.error` and returns only `{ authenticated: false }`. Clean.

- **Minor — "Fix N" comments**: Removed from `logs-tab.tsx`. Clean.

- **Minor — push failure log level**: Bumped from `debug` to `warning`. Good.

---

### ❌ Still outstanding

**Critical — `TOOL_CALL_RESULT` event type unverified, and `OperationalEventWriter` still has no unit tests**

`operational_events.py` is unchanged on both counts. The mapping `"TOOL_CALL_RESULT" → "tool_result"` is still present with `TOOL_CALL_END` in `_SKIP_TYPES`, and there are no pytest tests for `OperationalEventWriter` in the directory.

The test plan item "Runner pushes operational events visible in Logs tab (requires rebuilt runner image)" is checked, which is encouraging — but manual observation can't catch a silent regression where one event type name changes. The `TOOL_CALL_RESULT` naming question still needs one of:

1. A comment citing where `TOOL_CALL_RESULT` appears in the ag_ui library (e.g. `ag_ui.core.EventType.TOOL_CALL_RESULT`), **plus** at least a minimal pytest test for `consume()` that confirms the type string maps to `"tool_result"` — so future ag_ui upgrades don't silently break it.

---

### Minor cleanup (non-blocking, can be follow-up)

`SessionData` in `session.ts` still has `customTokenId?: string` — dead field now that `token-store.ts` is gone and nothing reads or writes it. Remove it to keep the type accurate.

---

One item to resolve, then this is mergeable.

markturansky (approved):
## Re-review — LGTM ✅

All items from my previous two reviews are resolved. Reviewing the two commits pushed after my last review:

### `cea4a275` (already in scope of re-review, now fully assessed)

The `operational_events.py` comment I flagged as missing is present: `# ag_ui.core.EventType.TOOL_CALL_RESULT is emitted by handlers.py:199`. I independently confirmed this in the upstream `ag-ui-protocol/ag-ui` repo — `TOOL_CALL_RESULT = "TOOL_CALL_RESULT"` is in `sdks/python/ag_ui/core/events.py`, and `handlers.py` for the claude-agent-sdk explicitly documents "Emits TOOL_CALL_END and TOOL_CALL_RESULT events."

`tests/test_operational_events.py` (7.9KB, 33 test cases) covers the full surface:
- All `_build_payload` branches including TOOL_CALL_RESULT truncation at 2000 chars
- `consume()` → confirms TOOL_CALL_RESULT maps to `"tool_result"` ✅
- Skip list behavior for TOOL_CALL_ARGS and TEXT_MESSAGE_CONTENT ✅
- None grpc_client and None event type edge cases ✅

### `0a099b7` — cookie split for overflow

The cookie overflow scenario (K8s SA tokens ~900 bytes raw → ~5.8KB encrypted, exceeding the 4KB browser cookie limit) is real and the two-cookie approach is the right fix:

- `ContextSessionData { customToken? }` is a clean, focused type
- `getContextSession()` inherits the same `SESSION_SECRET`, `httpOnly: true`, `sameSite: lax` — consistent security properties with the auth cookie
- `SessionData` no longer has the stale `customTokenId` field I flagged
- `setCustomContext` and `resetContext` correctly save both sessions
- The proxy route reads through `getRuntimeConfig()` unchanged — no call-site churn

**Minor, non-blocking (no action needed this PR):** `setCustomContext` saves the two cookies sequentially, not atomically — a failed `ctxSession.save()` leaves URL updated but token unset. This is an acceptable degradation: the user sees "custom context active" with no custom token, which gracefully falls back to SSO auth on the next request.

---

Approve. Ready to merge.

jsell-rh:
@Mergifyio queue

markturansky (commented):
# Amber Post-Merge Review — PR #1633

PR is merged. This is a retrospective pass to document what landed and flag two items that need follow-on work.

---

## What's clean ✓

- **`/api/me`**: error handler logs server-side and returns `{ authenticated: false }` — no raw error exposure. Correct.
- **`session.ts`**: split-cookie pattern (`ambient-ui-session` + `ambient-ui-session-ctx`) correctly solves the iron-session 4KB overflow. Tokens stay in encrypted httpOnly cookies, not process memory. Correct.
- **`auth.ts`**: oauth-proxy and none-auth modes removed cleanly; single native-SSO path is much simpler.
- **`operational_events.py`**: `OperationalEventWriter` is clean — async with `run_in_executor`, content truncated at 2000 chars, correct TOOL_CALL_RESULT mapping, exception logged and swallowed without crashing the fan-out loop.
- **`logs-tab.tsx`**: good a11y — `role="log"` on the event list, `aria-live` via `EventAnnouncer`, `aria-pressed` on filter chips, `aria-label` on the filter group.

---

## Follow-on work needed

### Major — `logs-tab.tsx`: raw `error.message` exposed in UI

```tsx
// logs-tab.tsx
if (error) {
  return (
    <div className="pt-4">
      <p className="text-sm text-destructive">
        Failed to load messages: {error.message}   {/* ← raw API error */}
      </p>
    </div>
  )
}
```

Whatever the fetch throws (network error text, HTTP status string, internal detail) is rendered directly. Matches the `raw-error-message` pattern from prior reviews.

**Fix:**
```tsx
<p className="text-sm text-destructive">
  Failed to load session events. Please try again.
</p>
```

Log the detail server-side if needed; don't surface it in the UI.

---

### Major — `use-session-messages.ts`: unconditional 3-second polling

```ts
// use-session-messages.ts
return useQuery({
  queryKey: queryKeys.messages.list(sessionId),
  queryFn: () => adapter.list(sessionId, { size: 1000 }),
  enabled: !!sessionId,
  refetchInterval: 3000,   // ← always polling, even for terminal sessions
})
```

The PR description says "Smart polling: stopped for terminal sessions" but `useSessionMessages` polls at 3s regardless of session phase. A Completed or Failed session will keep firing a GET every 3 seconds indefinitely as long as the Logs tab is open.

**Fix:** accept `session` (or `phase`) as a parameter and gate the interval:

```ts
const TERMINAL_PHASES = new Set(['Completed', 'Failed', 'Stopped'])

export function useSessionMessages(
  sessionId: string,
  phase?: string,
  port?: SessionMessagesPort,
) {
  const adapter = port ?? getDefaultPort()
  const isTerminal = phase ? TERMINAL_PHASES.has(phase) : false
  return useQuery({
    queryKey: queryKeys.messages.list(sessionId),
    queryFn: () => adapter.list(sessionId, { size: 1000 }),
    enabled: !!sessionId,
    refetchInterval: isTerminal ? false : 3000,
  })
}
```

The `LogsTab` already receives `session: DomainSession` and can forward `session.phase`.

---

## Minor

- **`auth.ts`**: `await import("@/lib/session")` is a dynamic lazy import with no reason to be lazy — static import is cleaner and avoids the async overhead on every request.
- **`use-session-messages.ts`**: `size: 1000` hardcoded — long-running sessions with many tool calls will return large payloads on every poll tick. Consider cursor-based pagination or streaming when the message count grows.

---

Both Major items should be tracked as follow-on issues before the Logs tab sees production traffic at scale.

markturansky:
## Amber Re-Review — PR #1633 (session log details tab)

*Post-merge review pass. Expanded from initial comment to cover proxy route, runtime-config, use-sessions, and pod_sync.*

---

### Two findings remain open (should be tracked as follow-on issues)

#### MAJOR — `useSessionMessages`: unconditional polling on terminal sessions

**File:** `components/ambient-ui/src/queries/use-session-messages.ts`

```typescript
export function useSessionMessages(sessionId: string, port?: SessionMessagesPort) {
  return useQuery({
    queryKey: queryKeys.messages.list(sessionId),
    queryFn: () => adapter.list(sessionId, { size: 1000 }),
    enabled: !!sessionId,
    refetchInterval: 3000,   // ← always polls, even for Completed/Failed sessions
  })
}
```

`useSessions` and `useSession` both implement phase-aware polling (1 s transitioning / 3 s active / `false` terminal) — confirmed in `use-sessions.ts`. `useSessionMessages` does not. Once a session reaches `Completed` or `Failed` the messages list is immutable; polling it every 3 s is wasted traffic.

**Fix:** accept a `phase` param (or read it from React Query cache) and set `refetchInterval: TERMINAL_PHASES.has(phase) ? false : 3000`.

---

#### MAJOR — `logs-tab.tsx`: raw `error.message` rendered in UI

**File:** `components/ambient-ui/src/app/(dashboard)/[projectId]/fleet/[sessionId]/_components/logs-tab.tsx`

```tsx
if (error) {
  return (
    <div className="pt-4">
      <p className="text-sm text-destructive">
        Failed to load messages: {error.message}   {/* ← raw error leaked to user */}
      </p>
    </div>
  )
}
```

`error.message` can contain network details, internal API paths, or server-side error strings. Convention is generic user-facing copy with detail logged server-side.

**Fix:**
```tsx
Failed to load session events. Please try again.
```

---

### Rest of the PR is clean

| Area | Verdict |
|---|---|
| `/api/me` — error handling | ✓ logs server-side, returns `{ authenticated: false }` |
| `session.ts` — split-cookie | ✓ both cookies correct; tokens encrypted, not in process memory |
| `runtime-config.ts` — custom token storage | ✓ reads from `getContextSession()` (iron-session), not env/memory |
| `[...path]/route.ts` — proxy | ✓ path traversal guard, SSE streaming, generic 502 on error |
| `pod_sync.go` — OOMKilled detection | ✓ `unstructured.Nested*` helpers, terminal-phase guard, restartCount threshold |
| `operational_events.py` — OperationalEventWriter | ✓ async executor, 2000-char truncation, exceptions swallowed after logging |

---

*Amber · background review agent · `3EaiLG7uLbZz3iwTGquqPPRTaCm`*

PR #1640, opened 2026-06-03
spec(security): RBAC runtime enforcement specification

jsell-rh:
## Summary

- Adds `specs/security/rbac-enforcement.spec.md` — defines how the ambient-api-server enforces scope-aware authorization at runtime
- Bridges the gap between the existing RBAC data model (roles, bindings, permissions) and actual access control
- 10 requirements with concrete Given/When/Then scenarios covering: scope-aware evaluation, list filtering, bootstrap, credential self-service, admin seeding, RoleBinding mutation guards, gRPC parity, service caller bypass, integration testing, and production rollout

## Key design decisions

- **Bootstrap via project creation** — `POST /projects` is auth-exempt; auto-creates `project:owner` binding atomically
- **User auto-provisioned from JWT** — no explicit registration, no permissions granted
- **Credential self-service** — `POST /credentials` auth-exempt; binding to projects requires `project:owner`
- **Scope-aware enforcement** — a binding for Project A cannot access Project B
- **List endpoints return filtered results, not 403** — empty list for zero access (matches K8s behavior)
- **Escalation prevention** — editors cannot grant owner; last-owner deletion returns 409
- **Admin seeded via CLI** — breaks the chicken-and-egg bootstrap problem
- **Rollout gated behind config flag** — no big-bang cutover

## Test plan

- [ ] Spec reviewed for format compliance against `specs/index.spec.md`
- [ ] Cross-referenced against existing specs: `ambient-model.spec.md`, `security.spec.md`, `sso-authentication.spec.md`
- [ ] No terminology collisions with Ambient domain model
- [ ] No implementation details in spec (no function names, file paths, pseudocode)

## Summary by CodeRabbit

* **Documentation**
  * Added RBAC enforcement specification documenting scope-aware authorization across the platform, including user provisioning flows, platform admin configuration, auth-exempt endpoints, permission filtering behavior, integration testing requirements, and production rollout considerations.


markturansky (changes requested):
## Review — `specs/security/rbac-enforcement.spec.md`

**Verdict: REQUEST CHANGES**

Cross-referenced against `security.spec.md`, `sso-authentication.spec.md`, and the PR's own internal consistency. The structure and coverage are solid — scope-aware evaluation, bootstrap pattern, service caller bypass, error opacity, and rollout gating are all well-reasoned. Two defects need to be corrected before this merges; several gaps should be closed at the same time.

---

### Major — Must Fix

#### 1. Privilege escalation rule is self-contradictory

The general rule in "Requirement: RoleBinding Mutation Authorization" states:

> A user MAY grant any role that is **at or below** the highest role they hold in the same scope.

But the spec immediately gives a concrete example that violates this rule:

> A `project:owner` on proj-1 MAY grant `project:editor` or `project:viewer` on proj-1, **but not `project:owner`**.

Under "at or below," `project:owner` is at the same level as itself, so the rule permits granting it. The scenario "Project editor cannot grant project owner" also relies on a "strictly below" interpretation — editors can't grant at their own level.

The intent (preventing escalation at the same tier) is correct, but the stated rule doesn't capture it. Two options:

- Change the rule to "**strictly below** the highest role they hold" and list `platform:admin` as the sole exception that may grant at its own level (since there's nothing above it).
- Or: define the hierarchy levels explicitly (e.g. level 0 = admin, level 1 = owner, level 2 = editor, ...) and state that a user may only grant roles at a strictly higher level number.

Either way, reconcile the rule with the `platform:admin → platform:admin` scenario, which **should** be permitted and currently contradicts the fix if you go with "strictly below" without the exception.

#### 2. `credential:reader` vs. `credential:viewer` naming collision

The role hierarchy in this spec uses **`credential:reader`**:

> `credential:owner` > `credential:reader` > `credential:token-reader`

`security.spec.md` defines **`credential:viewer`** (same role, different name):

> Credential CRUD governed by `credential:owner` role. Users with `credential:owner` can create, update, and delete credentials they own and bind them to Projects where they hold `project:owner`. Users with **`credential:viewer`** can read metadata (not tokens) on credentials bound to Projects they have access to.

Pick one name and make the specs agree. This inconsistency will cause divergent implementations. I'd suggest `credential:viewer` to be consistent with the `project:viewer` and `agent:observer`/`platform:viewer` pattern already established in `security.spec.md`.

---

### Notable Gaps — Address in This PR

#### 3. `credential:token-reader` must be marked platform-internal in the escalation table

`security.spec.md` is explicit: "`credential:token-reader` SHALL be granted to the runner service account by the platform at session start. It MUST NOT be granted via user-facing `POST /role_bindings`."

The escalation table in this spec lists `credential:token-reader` at the bottom of the credential hierarchy — which implies a `credential:owner` could grant it to any user. This contradicts the security spec's hard prohibition. The escalation requirement should call out that `credential:token-reader` is a platform-internal binding, not user-grantable, and attempting to grant it via `POST /role_bindings` returns 403.

#### 4. `agent:runner` role is missing from the escalation hierarchy

`security.spec.md` defines an `agent:runner` role ("Pods get minimum viable credential: read agent definition, push session messages, send inbox"). This spec's agent role hierarchy only lists `agent:operator` > `agent:observer`. Where does `agent:runner` fit? If it's below `agent:observer` and also platform-internal (like `credential:token-reader`), say so. If it's been renamed, the other spec needs updating too.

#### 5. Last-owner protection is defined for projects but not credentials

The spec correctly prevents orphaned projects with the 409 Conflict on last-owner deletion. The same problem exists for credentials: if the sole `credential:owner` deletes their binding, the credential becomes unmanageable (no one can update, delete, or rebind it). Either:

- Apply the same 409 protection to `credential:owner` deletions, or
- Explicitly state this is out of scope and explain the intended recovery path (e.g., platform admin cleanup via CLI).

#### 6. gRPC unauthorized idle streams have no resource boundary

For unauthorized `WatchSessions` streams the spec says "the stream remains open but idle (no error for watches)." An external caller with a valid JWT but zero bindings can open an unlimited number of idle watch streams and hold them indefinitely. The spec should define either a max-idle timeout or a max-streams-per-user limit to prevent connection exhaustion. Without this, the behavior is a DoS surface.

---

### Minor — Clarify if Easy, Defer if Not

**Auto-provisioning and service callers**: The auto-provisioning requirement says "when a JWT-authenticated caller is seen for the first time, a User record is created." Service callers authenticate with a service token, not a JWT. Add a sentence confirming that auto-provisioning does not trigger for service callers — it is scoped to JWT-authenticated human users only. This eliminates ambiguity at implementation time.

---

### What's Working Well

- Scope hierarchy and union-only model are precisely specified with scenarios that cover the key cross-scope isolation cases.
- Atomic `project:owner` binding on project creation is the right call and the rollback scenario is explicitly called out.
- 200+empty (not 403) for list endpoints matches K8s behavior and is explicitly defended in the design decisions table.
- Error opacity requirement is appropriately security-conscious.
- The service caller bypass rationale is clear and the two scenarios (HTTP + gRPC) cover the important cases.
- Config-flag rollout with explicit sequencing steps (seed → verify staging → migrate users → enable) is a safe operational model.
- No implementation detail (no function names, file paths, or pseudocode) leaks into the spec — format is clean.

Fix items 1 and 2 to unblock; 3–6 are serious enough that they should be addressed in this PR rather than deferred, since they affect the contract that implementation will be built against.

markturansky (changes requested):
## Re-review — `specs/security/rbac-enforcement.spec.md`

**Verdict: REQUEST CHANGES — one remaining issue**

All six findings from the first review are resolved. The revision is thorough: the level table eliminates the escalation rule ambiguity, `credential:viewer` is consistent with `security.spec.md`, platform-internal roles are explicitly blocked from user-granting, last-owner protection covers both projects and credentials, gRPC idle streams have a resource bound, and service caller auto-provisioning is clarified. Good work on the round 2 fixes.

One new cross-spec inconsistency found that needs to close before LGTM.

---

### Remaining Issue — Credential binding requires BOTH `credential:owner` AND `project:owner`

The normative text in "Credential Self-Service Bootstrap" currently states:

> Binding a credential to a project (`POST /role_bindings` with `scope=credential`) SHALL require the caller to hold `project:owner` on the target project.

`security.spec.md` is explicitly more restrictive:

> Users with `credential:owner` can create, update, and delete credentials they own and **bind them to Projects where they hold `project:owner`**.

The existing spec requires **both** conditions. This spec only enforces one. A project owner who knows a credential's UUID — but does not own it — could bind someone else's credential to their project, giving their session runners access to a credential they were never authorized on. Credential list filtering prevents discovery in the normal case but is not a substitute for authorization on the binding write path.

The scenario "Credential owner binds to their project" already models the correct dual-check (user A owns credential C AND holds `project:owner` on proj-1), but the normative requirement text doesn't enforce both. The scenario and the requirement contradict each other — the requirement text is the weaker one.

**Fix:** Update the normative text to:

> Binding a credential to a project (`POST /role_bindings` with `scope=credential`) SHALL require the caller to hold **`credential:owner` on the credential being bound AND `project:owner` on the target project**.

Also add a scenario confirming that a project owner who does not own the credential is rejected:

> - GIVEN user B holds `project:owner` on proj-1 but does NOT hold `credential:owner` on credential C
> - WHEN user B calls `POST /role_bindings` with `scope=credential`, `credential_id=C`, `project_id=proj-1`
> - THEN the request returns 403 Forbidden

---

### Confirmed Fixed

- Escalation rule: "strictly below" with level table and `platform:admin` exception — ✓
- `credential:viewer` naming consistent with `security.spec.md` — ✓
- `agent:runner` and `credential:token-reader` marked platform-internal, non-user-grantable — ✓
- Last-owner 409 protection extended to credentials — ✓
- gRPC idle stream resource limit specified — ✓
- Auto-provisioning scoped to JWT-authenticated human users only — ✓

Fix the credential binding authorization requirement and this is ready.

— amber

markturansky (commented):
## Re-review — round 3 — LGTM ✅

The round 2 finding is fixed. The normative text in "Credential Self-Service Bootstrap" now requires **both** `credential:owner` on the credential AND `project:owner` on the target project, consistent with `security.spec.md`. The new scenario "Non-credential-owner cannot bind credential to project" covers the previously-missing enforcement case. The rename of the existing scenario to "Non-project-owner cannot bind credential to project" is a nice precision improvement.

All findings from rounds 1 and 2 are confirmed resolved. The spec is internally consistent and cross-spec consistent with `security.spec.md` and `sso-authentication.spec.md`.

**Non-blocking nit for follow-up:** The design decisions table entry for credential creation still reads "Binding credentials to projects requires `project:owner`, preventing unauthorized sharing" — the rationale should match the updated requirement ("both `credential:owner` and `project:owner`"). Worth a one-line fix in a follow-up commit or after merge.

Ready to merge once the draft status is lifted.

— amber

markturansky (changes requested):
## Security-focused re-review — REQUEST CHANGES

Supersedes the round 3 LGTM. On a deeper pass with a security-specific lens, four issues surface that shouldn't ship in this spec. Two are authorization bypass/escalation paths; two are operational risks that could cause a live incident on rollout.

---

### Critical — Must Fix

#### 1. 403 on inaccessible resources leaks resource existence (ID enumeration)

"Error Response Opacity" says the middleware SHALL NOT disclose which permissions are missing. But the spec doesn't address the HTTP 403 vs 404 distinction for resource existence. If `GET /projects/proj-X` returns 403 for a caller without access, it tells that caller that proj-X exists. If it returns 404, it reveals nothing. The difference is exploitable: an attacker can enumerate valid project, agent, and session IDs by probing for 403 vs 404 responses.

The spec should state:

> For singleton resource endpoints (`GET /projects/{id}`, `GET /agents/{id}`, etc.), the middleware SHALL return 404 when the caller has no binding that covers the requested resource — regardless of whether the resource actually exists. Returning 403 leaks resource existence.

The "New user cannot access existing resources" scenario currently says "the project get returns 403." This is the wrong behavior under a strict opacity requirement and should be corrected to 404.

Note that list endpoints correctly return 200+empty (not 403), so the existing filtering behavior is fine. The problem is only singleton GETs.

#### 2. Credential-scoped role grants have no scope authorization check defined

The escalation rule defines a level hierarchy (strictly below own level). The scope check defines that a `project:owner` on proj-1 cannot grant on proj-2. But neither rule covers this case:

> User B has `project:owner` on proj-1. User B calls `POST /role_bindings` with `role=credential:viewer`, `credential_id=C`, `user_id=X`.

Under the level hierarchy, `project:owner` is Level 1 and `credential:viewer` is Level 2 — strictly below, so the escalation check passes. Under the scope check, proj-1 ownership is not involved since the new binding is credential-scoped — so that check doesn't block it either. User B can grant `credential:viewer` on a credential they don't own.

The spec needs a rule for credential-scoped role grants: the caller must hold `credential:owner` on the target credential (in addition to passing the level check). Without this, any authenticated user who has been a project owner can grant read access to credentials they don't own — including credentials owned by other users.

Suggested addition to "Requirement: RoleBinding Mutation Authorization":

> For credential-scoped role bindings (`scope=credential`), the caller SHALL hold `credential:owner` on the target credential in addition to satisfying the level hierarchy check.

Add a scenario to confirm:

> - GIVEN user B holds `project:owner` on proj-1 but does NOT hold `credential:owner` on credential C
> - WHEN user B calls `POST /role_bindings` with `role=credential:viewer`, `credential_id=C`, `user_id=X`
> - THEN the request returns 403 Forbidden

---

### High — Address Before Merge

#### 3. `credential:owner` binding is not specified as atomic with credential creation

The spec explicitly requires that the `project:owner` binding be created in the same database transaction as the project:

> "The RoleBinding is created in the same database transaction — if the RoleBinding creation fails, the project creation is rolled back."

No equivalent atomicity requirement exists for credential creation. If a credential is persisted but its `credential:owner` binding fails, the credential is immediately orphaned — no user holds `credential:owner`, no one can update, delete, rebind, or delete it, and the 409 protection means even platform admin cannot remove the phantom last-owner binding (there is no last-owner binding). Recovery requires direct database access.

Add a scenario mirroring the project atomicity one:

> - GIVEN a user calls `POST /credentials`
> - WHEN the credential is persisted
> - THEN the `credential:owner` RoleBinding is created in the same database transaction
> - AND if the RoleBinding creation fails, the credential creation is rolled back

#### 4. `project:owner` binding creation during enforcement-disabled rollout is unspecified

"Production Rollout" says enforcement is gated behind a config flag: when disabled, "RBAC is not evaluated AND the request proceeds." The spec does not say whether `POST /projects` creates the `project:owner` RoleBinding when enforcement is disabled.

If enforcement is disabled and the binding is NOT created, every project created during the rollout window will be orphaned the moment enforcement is enabled — no owner, no one can administer it, the 409 protection prevents cleanup. This is a silent incident trigger.

Add a normative statement:

> The `project:owner` RoleBinding SHALL be created on `POST /projects` regardless of whether RBAC enforcement is enabled. The binding creation is not gated by the enforcement flag, only authorization evaluation is.

Same applies to `credential:owner` on `POST /credentials`.

---

### Notable Gaps — Address If Possible, Defer If Not

#### 5. Concurrent first-time user provisioning is not idempotent by spec

The "Existing user not duplicated" scenario covers the sequential case (second request finds an existing record). A user making two simultaneous first-time requests could race to create the User record — both paths check "no record exists" before the first insert commits. The spec should state the creation SHALL be idempotent (e.g., implemented as an upsert keyed on `sub`). A non-idempotent implementation creates duplicate records or fails one of the concurrent requests with an unhandled constraint violation.

#### 6. Service token bypass has no cluster-internal protection stated

"Service callers are identified by authenticating with the platform service token." A stolen service token grants full RBAC bypass — there is no scope limitation on what a service caller can do. `security.spec.md` already requires that runner-internal endpoints "MUST validate the caller is cluster-internal to prevent token exfiltration." This spec should either reference that protection or explicitly state that the service token endpoint SHALL only be reachable from within the cluster, so exfiltration of the service token does not automatically grant external access.

---

### Prior Findings Confirmed Resolved

- Escalation rule: "strictly below" with level table and `platform:admin` exception — ✓
- `credential:viewer` naming consistent with `security.spec.md` — ✓
- Platform-internal roles (`agent:runner`, `credential:token-reader`) blocked from user grants — ✓
- Last-owner 409 for projects and credentials — ✓
- gRPC idle stream resource limit — ✓
- Auto-provisioning scoped to JWT users only — ✓
- Credential binding dual-check (`credential:owner` AND `project:owner`) — ✓

Items 1 and 2 above are authorization correctness issues that would ship as exploitable bugs. Items 3 and 4 are operational correctness issues that would cause an incident on rollout. All four should be in the spec before this merges.

— amber

markturansky (approved):
## Re-review — round 5 — LGTM ✅

All four round-4 required fixes are in, and both notable gaps from that pass are addressed as well. The spec is now security-complete.

### Confirmed Fixed

**Critical:**

1. **403 → 404 on unauthorized singleton GETs** — "Error Response Opacity" now explicitly requires 404 for singleton GETs with no binding, distinguishing correctly from mutation endpoints (which retain 403 for resources the caller can see). The "New user cannot access existing resources" scenario corrected to "returns 404 (existence not disclosed)". New scenario "Singleton GET returns 404, not 403" covers the case normatively. Design decisions entry explains the ID-enumeration rationale. ✓

2. **Credential-scoped grant scope check** — "RoleBinding Mutation Authorization" now states: "For credential-scoped role bindings (`scope=credential`), the caller SHALL hold `credential:owner` on the target credential in addition to satisfying the level hierarchy check." New scenario "Non-credential-owner cannot grant credential-scoped roles" (project:owner without credential:owner returns 403). Design decisions entry added. ✓

**High:**

3. **Credential owner binding atomicity** — New scenario "Credential owner binding is atomic with creation" mirrors the existing project atomicity scenario exactly: same database transaction, rollback on failure. ✓

4. **Owner binding creation decoupled from enforcement flag** — Production Rollout now normatively states: "Binding creation is not gated by the enforcement flag — only authorization evaluation is." New scenario "Bindings created even when enforcement is disabled" covers the rollout window case with explicit assertion that the binding is created before enforcement is turned on. Design decisions entry explains the orphan-prevention rationale. ✓

**Notable gaps (also fixed):**

5. **Concurrent first-time user provisioning** — New scenario "Concurrent first-time requests are idempotent" specifies upsert semantics keyed on identity claim. Design decisions entry added. ✓

6. **Service token cluster-internal protection** — Service Caller Bypass now requires the service token endpoint to be reachable only from within the cluster, with cross-reference to `security.spec.md#requirement-proxy-authentication`. New scenario "Service token rejected from outside the cluster" covers the exfiltration case. Design decisions entry added. ✓

### All Prior Findings Confirmed Resolved

- Escalation rule: "strictly below" with level table and `platform:admin` exception — ✓
- `credential:viewer` naming consistent with `security.spec.md` — ✓
- Platform-internal roles (`agent:runner`, `credential:token-reader`) blocked from user grants — ✓
- Last-owner 409 for projects and credentials — ✓
- gRPC idle stream resource limit — ✓
- Auto-provisioning scoped to JWT users only — ✓
- Credential binding dual-check (`credential:owner` AND `project:owner`) — ✓

### Summary

The spec is internally consistent, cross-spec consistent with `security.spec.md` and `sso-authentication.spec.md`, and covers the security surface area (resource enumeration, privilege escalation, credential ownership, atomicity, rollout safety, service token blast radius) at the level of detail needed to implement correctly. Ready to merge once the draft status is lifted.

— amber

markturansky:
# Review: PR #1640 — RBAC Enforcement Spec

**Score: 3/5 checks passed.** Strong conceptual work — the bootstrap, escalation prevention, and service-caller bypass design are solid. But there are two blockers and three criticals that need resolution before this spec can safely guide implementation.

---

## Blockers

### B1 — Mutation endpoints defeat the opacity requirement

The "Error Response Opacity" requirement correctly defines GET singleton → 404 when caller has no binding. But the "Forbidden response on mutation is opaque" scenario applies only to a caller who can **already see** the resource ("user A lacks permission to mutate a resource **they can see**"). It returns 403 for that case.

That leaves undefined: what does `PATCH /projects/proj-1` return when the caller has **zero bindings** (GET would return 404)?

If mutation endpoints return 403 for the no-binding case, an attacker can distinguish:
- `DELETE /projects/proj-1` → 404 = resource doesn't exist (or attacker has no binding)
- `DELETE /projects/proj-1` → 403 = resource exists but viewer access isn't enough

This breaks the existence-hiding invariant. The opacity requirement must be extended:

> For all resource-scoped endpoints (GET, PATCH, PUT, DELETE), when the caller has no binding covering the resource, the response SHALL be 404 — regardless of whether the resource exists or what the request method is.

The 403 response belongs only when the caller has a binding that grants read access but not the requested mutation.

### B2 — Last `platform:admin` deletion is undefined

The spec defines 409 for deleting the last `project:owner` binding and the last `credential:owner` binding. But it doesn't cover the last `platform:admin` case.

Deleting the last `platform:admin` would leave the platform without any admin — no one can seed a new one via the API, and the CLI-seeding escape hatch is an out-of-band operation. This must be specified:

> The system SHALL return 409 Conflict when a `DELETE /role_bindings` request would remove the last `platform:admin` binding on the platform.

---

## Critical

### C1 — Relationship to K8s RBAC (SSO roadmap) is undefined

The SSO authentication spec describes the current model: K8s impersonation-based authorization, with "Phase 4. DB RBAC reconciler" as a future state where "DB `role_bindings` table becomes the single write plane." This spec appears to describe that Phase 4 target state — but it doesn't say so, and it doesn't acknowledge the K8s RBAC model that's live in production today.

During the transition window:
- Which system gates access? Both? DB RBAC only when the flag is on?
- Is the enforcement flag in this spec the same flag as the SSO spec's feature flag, or a separate one?
- What happens when a user has a K8s RoleBinding but no DB RoleBinding (or vice versa) during coexistence?

This spec needs a `## Scope` or `## Relationship to Existing Authorization` section that places it in the IAM roadmap, defines what it supersedes, and specifies coexistence behavior during rollout.

### C2 — `GET /role_bindings` access model is undefined

The spec defines who can mutate role bindings (escalation rules) and who can delete them (project/credential owners). It never defines who can **list or read** role bindings.

This is a security concern: role bindings expose the full access topology of the platform. A project:viewer listing all role bindings would learn every user's access level across every project. Must specify:

- What permission is required to call `GET /role_bindings` and `GET /role_bindings/{id}`?
- Are list results filtered (a project:owner sees only bindings within their projects)?
- Can any authenticated user enumerate all platform:admin assignments?

### C3 — PATCH/DELETE of role bindings scope validation is undefined

The "RoleBinding Mutation Authorization" requirement covers `POST /role_bindings` (create) and one `DELETE` scenario (project:owner revoking a project:viewer). It doesn't define the authorization model for deleting bindings in general.

Gaps:
- Can a `project:owner` delete a `credential:owner` binding in their project? (They shouldn't — credential ownership is independent of project ownership.)
- What's the authorization check on `DELETE /role_bindings/{id}` beyond "is the caller's scope sufficient"?
- Does updating a binding via PUT/PATCH exist? If so, what's the authorization model?

---

## Major

### M1 — `agent:operator` and `credential:viewer` escalation edge cases have no scenarios

The hierarchy table places `agent:operator` and `credential:viewer` at Level 2. The "strictly below" rule implies:

- A `project:editor` (Level 2) **cannot** grant `credential:viewer` (same level)
- An `agent:operator` (Level 2) **can** grant `agent:observer` (Level 3) on the same agent

Both are probably intended but neither has a scenario. The `credential:viewer` case is particularly important: if a project:editor tried to share credential read access, the spec currently doesn't explicitly block it — it has to be inferred from the level table. Add explicit scenarios for both.

### M2 — Platform admin recovery for orphaned resources is unspecified

The Design Decisions table says: "Platform admins can always intervene as a recovery path" for orphaned projects/credentials. But no requirement or scenario defines what this means:

- Can a platform:admin delete the last `project:owner` binding (overriding the 409)?
- Can a platform:admin assume ownership of a project with no remaining owner?
- What's the explicit mechanism for admin intervention?

This is critical enough to warrant a formal requirement. "Platform admins can always intervene" in prose is not a spec — it's a promise without enforcement semantics.

### M3 — Integration test service-caller bypass undermines RBAC coverage

The integration test requirement says "Tests SHALL NOT disable the authorization middleware." But the service caller bypass (platform service token → full RBAC bypass) is a back door. Test infrastructure commonly uses service tokens for convenience. The spec should make this explicit:

> Test code that exercises user-facing endpoints SHALL authenticate with user JWTs and explicit role bindings. Service caller tokens MAY be used only for test setup operations (e.g., seeding test data via platform-internal routes). User-facing RBAC enforcement SHALL NOT be tested using the service caller bypass.

---

## Minor

### m1 — Idle gRPC stream scenario is out of scope

The "Idle watch stream resource limit" scenario belongs in a reliability or resource management spec, not an RBAC enforcement spec. It's about server-side connection management, not authorization. Extract or reference from here.

### m2 — "Proxy routes out of scope" limitation is buried

The design decisions note that proxy routes to external backends are outside the scope of this spec. This is an important boundary that implementers need to know upfront — move it to the `## Purpose` section as a formal scope exclusion.

---

## What's well-specified

- Scope-aware evaluation and the union-only model are clearly defined with good scenario coverage
- The bootstrap mechanics (project creation → atomic owner binding, admin CLI seeding) are watertight
- Credential self-service (dual ownership check: `credential:owner` AND `project:owner`) is correctly specified
- The escalation prevention logic (strictly-below, platform:admin exception, platform-internal roles blocked) is clean
- List filtering returning 200+empty vs 403 is correctly specified and well-motivated
- Service token restricted to cluster-internal traffic is a good security boundary
- Rollout flag + binding creation during disabled mode prevents orphaning — thoughtful

---

## Summary

Two blockers to fix before this guides implementation: the mutation-endpoint existence leak (B1) and the missing last-admin protection (B2). Three criticals for spec completeness: the K8s RBAC coexistence story (C1), the undefined role-binding read access model (C2), and the scope validation for binding deletes (C3). The majors (M1-M3) can be addressed in a follow-up or in the same revision — no blocker to merging once B1-B2 and C1-C3 are resolved.

---
*Amber · code review agent · 2026-06-04*

PR #1648, opened 2026-06-03
feat(spec): add AgenticApplication to data model spec

markturansky:
## Summary

- Adds **AgenticApplication** to `docs/internal/design/ambient-model.spec.md` — a GitOps continuous sync resource (Argo CD for Ambient) that reconciles agent fleet definitions from a git repository to a target Ambient instance
- Defines the full concept mapping from Argo CD (Source, Destination, Sync Status, Health, Self-Heal, Prune) to Ambient equivalents
- Clarifies that **Cluster and Ambient are infrastructure inventory (destinations), not sync targets** — AgenticApplication syncs Projects, Agents, RoleBindings, and Inbox seeds only
- Adds ERD entity, field reference table, sync lifecycle, destination resolution, multi-environment promotion examples
- Adds API endpoints (CRUD + `/sync`, `/refresh`, `/status` sub-resources)
- Adds CLI commands (`acpctl get/create/sync/refresh agentic-application`)
- Adds `gitops:admin` and `gitops:viewer` RBAC roles with permission matrix column
- Adds 6 design decisions and 3 implementation coverage matrix rows

## Design Highlights

 < /dev/null |  Argo CD | Ambient |
|---|---|
| Application | AgenticApplication |
| Cluster + Namespace | Ambient instance + Project |
| Helm / Kustomize / Jsonnet | Kustomize only (built-in engine from `acpctl apply -k`) |
| K8s API targeting | Ambient REST API targeting (SDK client, not kubeconfig) |
| App of Apps | Multiple AgenticApplications per sector (s0→s5 promotion) |

## Test plan

- [ ] Spec review: verify ERD relationships render correctly in Mermaid
- [ ] Verify field descriptions are complete and unambiguous
- [ ] Validate RBAC matrix is consistent with existing roles
- [ ] Confirm API endpoint naming follows existing conventions (`/api/ambient/v1/agentic_applications`)

## Summary by CodeRabbit

* **Documentation**
  * Added Application (GitOps) as a first-class entity with ER diagram and fields for git source, destination, credentials, sync/prune/self-heal flags, status, per-resource breakdowns, and audit timestamps.
  * New "Application — GitOps Continuous Sync" section detailing kinds synced, lifecycle (refresh/render/diff/sync/status), destination resolution, pruning rules, and unsupported-kind handling.
  * Extended API/CLI docs with CRUD plus /sync, /refresh, /status endpoints and status examples.
  * Added gitops:admin and gitops:viewer roles, updated permission matrix, labels/annotations conventions, design notes, and implementation coverage placeholders.

### DIFF
--- a/specs/api/ambient-model.spec.md
+++ b/specs/api/ambient-model.spec.md
@@ -2,7 +2,8 @@
 
 **Date:** 2026-03-20
 **Status:** Active
-**Last Updated:** 2026-05-12 — migrate Credentials from project-scoped to global routes (`/credentials`); remove `project_id` from model, OpenAPI, and SDK; add drop-column migration; update coverage matrix
+**Last Updated:** 2026-06-03 — added Application (GitOps continuous sync for agent fleets)
+**Previous:** 2026-05-12 — migrate Credentials from project-scoped to global routes (`/credentials`); remove `project_id` from model, OpenAPI, and SDK; add drop-column migration; update coverage matrix
 **Workflow:** `../../workflows/sessions/ambient-model.workflow.md` — implementation waves, gap table, build commands, run log
 **Design:** `credentials-session.md` — full Credential Kind design spec and rationale
 
@@ -19,6 +20,7 @@ The Ambient API server provides a coordination layer for orchestrating fleets of
 - **Inbox** — a persistent message queue on an Agent. Messages survive across sessions and are drained into the start context at the next run.
 - **Credential** — a global secret. Stores a Personal Access Token or equivalent for an external provider (GitHub, GitLab, Jira, Google, Vertex AI, Kubeconfig). Consumed by runners at session start. Bound to Projects via RoleBindings — a single Credential can be shared across multiple Projects without duplication.
 - **RoleBinding** — binds a Role to a subject (user or project) at a given scope. Ownership and access for all Kinds is expressed through RoleBindings. The subject and scope are each represented as typed nullable FKs — exactly one FK is non-null, determined by `scope`.
+- **Application** — a GitOps binding that continuously syncs agent fleet definitions from a git repository to an Ambient instance. The Ambient equivalent of an Argo CD Application.
 
 The stable address of an agent is `{project_name}/{agent_name}`. It holds the inbox and links to the active session.
 
@@ -224,6 +226,36 @@ erDiagram
         time   deleted_at
     }
 
+    %% ── Application (GitOps sync — Argo CD for Ambient) ──────────────
+
+    Application {
+        string ID PK "KSUID"
+        string name "unique; human-readable"
+        string source_repo_url "git repository URL"
+        string source_target_revision "branch, tag, or commit SHA"
+        string source_path "path within repo to kustomize overlay"
+        string destination_ambient_url "nullable — target Ambient API URL; null = local"
+        string destination_project "target project name; created if CreateProject=true"
+        bool   auto_sync "enable automated sync on git change"
+        bool   auto_prune "delete resources removed from git"
+        bool   self_heal "re-sync when live state drifts"
+        string sync_options "comma-separated: CreateProject=true, etc."
+        int    retry_limit "max sync retries on failure"
+        string sync_status "Synced | OutOfSync | Unknown"
+        string health_status "Healthy | Degraded | Progressing | Unknown"
+        string sync_revision "last successfully synced git commit SHA"
+        string operation_phase "Succeeded | Failed | Running | idle"
+        string operation_message "human-readable sync result summary"
+        jsonb  resource_status "per-resource sync/health detail"
+        jsonb  conditions "error conditions array"
+        jsonb  labels
+        jsonb  annotations
+        time   last_synced_at "timestamp of last successful sync"
+        time   created_at
+        time   updated_at
+        time   deleted_at
+    }
+
     %% ── Relationships ────────────────────────────────────────────────────────
 
     Project         ||--o{ ProjectSettings  : "has"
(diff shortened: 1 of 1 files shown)


markturansky (commented):
## Spec review — AgenticApplication data model

Solid foundational spec. The Argo CD analogy is well-constructed, the ERD is clean, and the "what gets synced / what does not" distinction between fleet definitions and infrastructure inventory is well-argued. Several gaps need closing before implementation starts.

---

### [SECURITY] RoleBinding sync bypasses escalation rules

The What Gets Synced table includes:

> `RoleBinding` | Created if not present; idempotent by user+role+scope key

The rbac-enforcement spec (merged) establishes strict escalation rules: a principal can only grant roles at levels strictly below their own. Those rules bind human API callers — but the AgenticApplication sync controller is acting as a service principal against the destination Ambient's REST API. If the controller has a powerful service credential, it can create RoleBindings that no human caller would be permitted to create.

Concretely: whoever controls the source git repo controls the RBAC of the destination project. A malicious or compromised repo commit could grant `project:owner` or `gitops:admin` to an arbitrary user.

Options to address:
1. **Restrict sync scope** — only allow the sync engine to create RoleBindings where the role is at or below `agent:observer` in the hierarchy; any binding above that requires human approval (manual sync + approval gate).
2. **Bind the sync engine by escalation rules** — state explicitly that the sync engine must satisfy the same escalation rules as the service credential it uses. This means the service credential level sets the ceiling on what RoleBindings the sync can create.
3. **Exclude RoleBindings from auto-sync** — mark RoleBindings as "manual review required" even when `auto_sync=true`.

At minimum, add a design decision that names this risk explicitly.

---

### [SECURITY] Remote Ambient authentication is not implementable as specified

> authentication via forwarded token or service credential

`AgenticApplication` has no field for a credential. When `destination_ambient_id` is set, the sync engine needs a token for the remote Ambient's REST API. The existing security spec defines a `Credential` type (with encrypted storage and credential:owner RBAC). The correct field is something like `credential_id FK → Credential.ID` — the stored credential provides the auth context for the remote Ambient SDK client.

"Forwarded token" is also dangerous: if the sync runs as an async background controller (as implied by `auto_sync` polling), there is no request context to forward a token from. The controller must use a stored credential.

Add a `credential_id` field to `AgenticApplication` and a design decision explaining how the credential is resolved (same `kubeconfig_enc`-style write-only pattern as Cluster?).

---

### [IMPORTANT] `gitops:admin` / `gitops:viewer` not integrated with escalation rules

Two new built-in roles are introduced but the escalation rule chain is not updated. The rbac-enforcement spec defines who can grant `project:owner`, `cluster:admin`, etc. Missing:

- Can `platform:admin` grant `gitops:admin`? (Almost certainly yes — but unstated.)
- Can `project:owner` grant `gitops:admin`? (Probably no — but unstated.)
- At what scope can `gitops:admin` be granted? Global only? Or project-scoped?
- Is `gitops:viewer` user-grantable or platform-internal?

Per the escalation spec pattern, platform-scoped roles (`cluster:admin`, and apparently `gitops:admin`) should only be grantable by `platform:admin`. Add this to the rbac-enforcement spec or add a normative statement here that `gitops:admin` / `gitops:viewer` are grantable only by `platform:admin`.

---

### [IMPORTANT] Shared kustomize engine / Cluster + Ambient kind handling is underspecified

The spec states:

> The sync engine reuses the same kustomize rendering logic as `acpctl apply -k`. Extracted to a shared package.

But `acpctl apply` supports `Cluster` and `Ambient` as apply kinds, while the AgenticApplication sync explicitly excludes them ("not synced"). If a kustomize overlay contains a `Cluster` or `Ambient` manifest:

- Does the sync engine skip it silently?
- Does it warn in `operation_message`?
- Does it error and set `operation_phase: Failed`?

Implementors need a clear answer. The shared engine model implies the same code path, which makes the divergence a runtime choice. Recommend: explicitly state that `Cluster`/`Ambient` documents in a kustomize tree are **silently skipped** by the sync engine (not applied), with a per-resource status entry of `{"kind": "Ambient", "name": "...", "status": "Skipped", "message": "infrastructure inventory — not synced by AgenticApplication"}`.

---

### [IMPORTANT] `auto_prune=true` on a Project is destructive; spec doesn't warn

The prune behavior for Agents is expected. But if a git overlay removes the `Project` manifest itself and `auto_prune=true`, the sync would delete the project — cascading through all agents, their sessions, inbox messages, and session message history. This is irrecoverable.

Add an explicit design decision or warning box:

> **Pruning a Project is permanently destructive.** All Agents, Sessions, Inbox messages, and SessionMessages in the project are deleted. The spec recommends setting `auto_prune=false` at the Project level, or implementing a "tombstone" pattern (mark project for deletion, require manual confirmation) before implementing project prune.

At minimum, note this in the `auto_prune` field description in the Field Reference.

---

### [NOTABLE] `project:owner` cannot manage GitOps sync for their own project

The permission matrix shows `project:owner` has `—` for AgenticApplications. This means a team that fully owns a project must escalate to a `gitops:admin` to create a GitOps binding for their own project — even though they already have full control over the project contents.

In Argo CD, the equivalent is that a namespace owner can create an Application targeting their own namespace (if allowed by the AppProject policy). The current design requires a platform-level role for what is effectively a project-level operation.

This may be intentional (AgenticApplications have cross-environment reach that exceeds project scope — the Ambient destination could be in another environment). If intentional, add it as an explicit design decision: "gitops:admin is platform-scoped because AgenticApplications can target any Ambient, including production environments; project:owner scope would not be sufficient to authorize cross-environment sync."

If not intentional, consider allowing project:owner to create AgenticApplications where `destination_project == their owned project AND destination_ambient_id == null` (local only), with `gitops:admin` required for remote Ambient targets.

---

### [NOTABLE] PUT vs PATCH inconsistency for Cluster and Ambient

Cluster and Ambient update endpoints are listed as `PUT`:
```
PUT /clusters/{id}
PUT /ambients/{id}
```

All other mutable resources use `PATCH` (partial update). PUT implies full replacement — callers must send the complete resource, not a diff. This is inconsistent with the pattern established for Projects, Agents, and AgenticApplications (all PATCH). Change to `PATCH /clusters/{id}` and `PATCH /ambients/{id}`.

---

### [MINOR] Inbox dedup key uses `from_name` which is mutable

> Inbox (seed messages) | Idempotent delivery — only new messages (by `from_name` + `body` dedup) are posted

`from_name` is "denormalized sender display name" — it can change if an agent is renamed. A seed message previously delivered under `from_name="lead"` would be re-delivered after an agent rename, breaking idempotency. Dedup by `from_agent_id + body` (where `from_agent_id` is the immutable FK) or by a content hash of the body alone.

---

### [MINOR] Health status "expected state" is underspecified

> `Healthy` = all agents exist and have expected state

"Expected state" needs definition. Candidates:
- Agent exists with exact prompt from git = Healthy
- Agent exists regardless of current prompt = Healthy (drift detected separately via `self_heal`)
- Agent exists and has an active session = Healthy

The third interpretation conflicts with the model ("only one session per agent; sessions are run artifacts"). Recommend the first or second. Define what triggers `Degraded` per resource kind.

---

### What's correct

- ERD relationships: AgenticApplication → Ambient (nullable FK), AgenticApplication → Project (syncs_to) are correct.
- "What does not get synced" is well-argued and clear.
- Sync lifecycle (Refresh → Render → Diff → Sync → Status) is the right sequence.
- Destination resolution (null → local Ambient) is clean.
- Multi-environment promotion via multiple AgenticApplications is idiomatic and matches Argo CD's App-per-env pattern.
- All AgenticApplication endpoints follow existing naming conventions (`/api/ambient/v1/agentic_applications`).
- `gitops:admin` not having RBAC column permissions in the matrix is correct (the role manages apps, not bindings).
- Design decisions section is comprehensive; the existing decisions are well-reasoned.
- Implementation coverage matrix rows for AgenticApplication correctly marked 🔲 planned throughout.

jsell-rh:
## Review: feat(spec): add AgenticApplication to data model spec

**TL;DR:** Pure spec PR. The core approach — modeling AgenticApplication as "Argo CD for Ambient" — is the right abstraction. The design decisions are largely sound. A handful of gaps worth addressing before merge.

---

### What's Good

- **Argo CD analogy is the right call.** The concept mapping table gives the team shared vocabulary without inventing new mental models.
- **Sync scope is deliberate.** Fleet definitions vs. infrastructure inventory is a clean boundary, well-justified.
- **Security-conscious design decisions.** RoleBinding escalation bounds, async credential model (stored FK, not forwarded token), project-prune manual confirmation — all correct for the threat model.
- **Multi-env via multiple AgenticApplications** is pragmatic. No "App of Apps" complexity.
- **Git polling v1 / webhooks v2** is the right phasing.

---

### Concerns

**1. Credential sync is a security gap (high priority)**

The spec lists `Credential` as a syncable kind, but doesn't address what happens to the token value on patch if a credential already exists with a different token. If the kustomize manifest carries a token in plaintext, that's secrets-in-git. The spec needs to explicitly state how token values are sourced during sync (env var injection? write-only on create, never updated on patch?).

**2. Private git repo auth is unspecified**

`source_repo_url` supports SSH and HTTPS but there's no mention of how private repo credentials are provided. Is this a separate `Credential` kind? An SSH key type? Most production repos are private — this is a significant gap.

**3. `sync_options` as a comma-separated string is a smell**

`"comma-separated: CreateProject=true, etc."` will be hard to parse and validate as options grow. Use a JSONB object or discrete boolean columns — the pattern is already established by `auto_sync`, `auto_prune`, `self_heal`. `CreateProject=true` should be a first-class field, not string parsing.

**4. Health vs. sync-correctness naming**

`health_status: Healthy` means "all synced resources match target state" — which is sync correctness, not runtime health. Agents don't have liveness/readiness like k8s pods. This naming will mislead. Consider `sync_health`, or explicitly call out in the spec that "health" here means "sync integrity, not runtime health."

**5. `project:owner` local-only access needs enforcement details**

The permission matrix says `project:owner` gets `local-only (own project)` for AgenticApps. The spec doesn't explain how "own project" is enforced at the API layer — is it checked against `destination_project`? What prevents a `project:owner` from setting `destination_project` to a project they don't own? Needs a design decision.

**6. No concurrency/locking spec**

If `auto_sync` fires while a sync is already `Running`, what happens — queue, skip, or cancel? `operation_phase: Running` captures state but controller behavior on overlap is undefined.

**7. `/status` endpoint vs. full resource appear redundant**

`GET /agentic_applications/{id}` is described as "includes status" and `GET /agentic_applications/{id}/status` exists as a separate endpoint returning the same fields. Clarify whether `/status` is a different projection or subset — otherwise it reads as duplication.

---

### Minor

- YAML example uses `credential: staging-ambient-pat` (name string) but the model field is `credential_id` (FK). Note that name resolution happens at the API layer, or use `credential_id: <uuid>` in the example.
- `AgenticApplication }o--o| Project : "syncs_to"` in the ERD — `destination_project` is a name, not a FK (intentional for cross-instance support). Worth a design decision noting the deliberate denormalization.
- `ScheduledSession` exclusion is hand-waved as "future sync candidate." It's a fleet definition kind, not ephemeral — the v1 exclusion reason should be stated explicitly.

---

### Verdict

Right approach, draft-quality gaps to close. Approve the direction; request addressing Credential token handling (security), private git repo auth, and `sync_options` string smell before this is implementation-ready.

markturansky (commented):
# Amber Review — PR #1648: AgenticApplication data model spec

Spec-only review. Living-doc convention is correct throughout — all implementation items are marked `🔲 planned`. Focused on internal consistency, completeness, and design gaps.

---

## Issues requiring fix before merge

### F1 — `operation_phase` value inconsistency: `"idle"` vs empty string

The ERD entity defines the field as:
```
string operation_phase "Succeeded | Failed | Running | idle"
```

The Field Reference table defines it as:
```
State of the last sync operation: Succeeded, Failed, Running, or empty if never synced
```

`"idle"` (ERD) and `""` (field ref) are different values. Pick one and make both sections consistent. Argo CD uses empty string for the not-yet-synced case — recommend aligning with that.

### F2 — ERD relationship implies FK that doesn't exist

The ERD includes:
```
AgenticApplication }o--o| Project : "syncs_to"
```

But `destination_project` is a **name string** (the field reference confirms: "Target project name"), not a `destination_project_id` FK. The ERD relationship implies a join column that doesn't exist. Either:
- Drop the `syncs_to` ERD relationship and replace with a comment noting it's by-name, not by FK, or
- Add a `destination_project_id` FK column (nullable) to the model

If it stays by-name: project renames will silently break sync — worth documenting explicitly in the field reference for `destination_project`.

### F3 — No mechanism for private git repo authentication

The field reference for `source_repo_url` says "HTTPS or SSH." But `credential_id` is documented as auth for the **destination Ambient API**, not the source git repository. There's no field for git authentication. This means:

- Public HTTPS repos: work
- Private HTTPS repos: no credential mechanism defined
- SSH repos: no key mechanism defined

Argo CD solves this with a separate Repository object. Options for this spec:
- Add `source_credential_id` FK → Credential (a Credential whose token is used as the git PAT / Basic auth for HTTPS, or holds an SSH private key for SSH)
- Restrict v1 to public HTTPS repos and mark SSH + private repo auth as `🔲 planned`

As written, advertising SSH support in the field reference is inaccurate since there's no auth mechanism for it.

---

## Non-blocking observations

### N1 — Agent auto-prune cascade to Sessions is undocumented

The `auto_prune` / Project prune safety section documents cascade behavior for Project deletion:
> "All Agents, Sessions, Inbox messages, and SessionMessages in the project are cascade-deleted"

But Agent-level auto-prune cascade is not described. If an Agent is deleted via `auto_prune`, what happens to its active sessions, inbox messages, and session messages? Worth one sentence in the auto_prune field reference.

### N2 — Polling interval is not a configurable field

The sync lifecycle section states "default: 3 minutes" for auto-sync polling. This implies a global constant, but it's not defined as a field on AgenticApplication. If the interval should be per-application (useful for high-frequency or low-priority fleets), it should be a field. If it's global config, note that explicitly. Currently, the spec implies it's configurable (the word "configurable") but provides no configuration surface.

### N3 — Concurrent sync semantics undefined

If `auto_sync=true` fires while a manual `POST /sync` is in `operation_phase: Running`, the spec doesn't define what happens. Does the auto-sync skip? Queue? Fail? Worth a single sentence in the Sync Lifecycle section.

### N4 — `health_status` conflates sync correctness with runtime health

The `health_status` definition: "Healthy = all synced resources exist in the destination and match the target state." This is really a sync-correctness check, not a runtime health check. An Agent could match git exactly but be in a failure state (all its sessions failing, no successful runs). Recommend documenting this explicitly: "For v1, health reflects sync correctness (resource exists and fields match git), not runtime agent health."

### N5 — `project:owner` self-service authorization logic not specified

The permission matrix and design decision both say `project:owner` can create local AgenticApplications for their own project. But the authorization enforcement isn't specified: what checks does the handler perform to validate `destination_ambient_url == null` AND `destination_project` is owned by the requester? The runtime enforcement spec should define this gate, similar to how `rbac-enforcement.spec.md` defines the B1–B5 requirements. Noting this as a gap for the implementation spec, not a blocker here.

---

## What's strong

**Project prune safety is correct.** The explicit block on auto-pruning a Project, and requiring `prune: true, prune_project: true` as dual opt-in flags on `POST /sync`, is the right design. Cascade destruction of a project requires intentional confirmation at every layer.

**RoleBinding escalation ceiling is well-specified.** The sync engine being bound by the credential's effective role level prevents a compromised git repo from escalating RBAC. This is the right invariant and it's clearly stated in both the sync table and the design decisions.

**Remote Ambient auth via no-cache Credential is correct.** Resolving the token fresh per sync cycle via `GET /credentials/{id}/token` and not caching it is the right pattern for async polling controllers. The design decision explaining *why* (no request context for async controllers) is the right level of documentation.

**Argo CD concept mapping is clear.** The table is accurate and useful for readers coming from an Argo CD background.

**Living-doc convention is correct.** All API, CLI, and coverage matrix rows are marked `🔲 planned`. The `> **Status:** Proposed` is not used here but isn't needed since the entity is being added to the model spec (not claiming implementation status).

---

Fix F1 and F2 before merging (small). F3 needs a design decision — either add the field or scope-limit the claim. N1–N5 are recommended for this PR or a follow-up.

*— Amber*

PR #1655, opened 2026-06-05
spec(security): credential token encryption at rest

jsell-rh:
## Summary
- Adds spec for AES-256-GCM encryption of credential tokens in PostgreSQL
- Versioned ciphertext format (`enc:v{N}:{base64}`) with credential ID as AAD
- Encryption key via K8s Secret → env var, with key rotation support
- CLI command (`encrypt-credentials`) for initial migration, re-encryption, rollback (`--decrypt`), and dry-run
- Graceful degradation for dev environments (plaintext mode with startup warning)
- Designed as stepping stone toward Vault-only storage — API contract unchanged

## Context
Credential tokens are currently stored as plaintext in PostgreSQL. The security spec claims "encrypted at rest" but no encryption exists. This spec closes that gap with a design that passed two rounds of critic review (schema, RBAC, consumer impact, cryptography).

## Test plan
- [x] Spec-only change — no code modifications
- [x] Two critic passes completed (4 critics each), all BLOCKERs/MAJORs resolved
- [ ] Implementation tracked separately

## Summary by CodeRabbit

* **Documentation**
  * Added a specification for encrypting credential tokens at rest: AES-256-GCM, versioned keys, and version-tagged ciphertext format.
  * Describes service-layer encryption/decryption, fail-closed startup behavior, plaintext opt-in, key rotation and bulk migration via CLI (with dry-run and rollback), binding credentials to ciphertext to prevent row-swapping, and required runtime settings.


jsell-rh:
@coderabbitai review

PR #1657, opened 2026-06-05
fix(api-server): address coderabbit review on credential encryption

jsell-rh:
## Summary
Follow-up to #1656 (auto-merged before review fixes could land). Addresses all 6 coderabbit findings:

- **CLI success counters**: track actual successes per bucket, not candidates-minus-failures
- **Duplicate key version rejection**: `"1"` and `"01"` normalizing to same int now errors
- **API token test assertion**: assert exact plaintext value, not just response shape
- **DAO error handling**: check errors in row-swap test instead of swallowing them
- **Fail-closed probe**: DB errors during encrypted-token check are fatal, not warnings
- **nil events guard**: `Replace` and `Delete` now guard `s.events` like `Create` does

## Test plan
- [x] `go test ./pkg/crypto/` — 20 tests pass
- [x] `go test ./plugins/credentials/` — 14 integration tests pass
- [x] `go build ./...` — compiles clean

## Summary by CodeRabbit

* **New Features**
  * Added detection of duplicate encryption key versions at startup

* **Bug Fixes**
  * Avoided nil-pointer issues when credential event handling is not configured
  * Made startup validation of encryption configuration fail fast on errors
  * Improved reporting to distinguish credentials that would be affected vs. those actually processed

* **Tests**
  * Added unit and integration tests for keyring behavior and token version handling

### DIFF
--- a/components/ambient-api-server/pkg/cmd/encrypt_credentials.go
+++ b/components/ambient-api-server/pkg/cmd/encrypt_credentials.go
@@ -79,7 +79,8 @@ func runEncrypt(db *gorm.DB, keyring *crypto.Keyring, activeVersion int, dryRun 
 		glog.Fatalf("Failed to query credentials: %v", err)
 	}
 
-	var plaintext, reencrypt, current, failed int
+	var plaintextCandidates, reencryptCandidates, current int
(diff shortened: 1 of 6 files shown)


jsell-rh:
@Mergifyio queue

PR #1689, opened 2026-06-15
fix(frontend): offset error toast to prevent overlaying stop button

squizzi:
This PR offsets the error toast to prevent it from annoyingly covering the stop button within a chat.  It also makes the toast dismissible so that if it does get in the users way they can close it.

### Before
Screenshot 2026-06-15 at 2 18 48 PM

### After
Screenshot 2026-06-15 at 2 15 16 PM

## Summary by CodeRabbit

* **Style**
  * Toast notifications are now positioned in the bottom-right corner with improved spacing and include a close button for user control.


markturansky:
LGTM. Tests pass.

